PR #3, opened 2017-10-11
KEP-38: Merge KEP-38-connect-to-eth-ropsten-network-and-get-token-count-for-given-eth-address-and-display-it-in-a-message-box-on-start to master

scottburch:


dmitrymukhin:
All changes from this PR also go into PR for KEP-40, no need to merge this. Keeping branch until the end of sprint just in case.

PR #4, opened 2017-10-13
KEP-40: Merge KEP-40-only-proceed-if-the-token-count-on-eth-network-for-given-address-is-at-least-100 to master

scottburch:


rnistuk:
We're not working on this project anymore, just closing the request

PR #5, opened 2017-11-08
[KEP-38] Getting token count from Ropsten network + unit test.

dmitrymukhin:
Added ethereum_api class and token_balance() method to pull balance for specific token on specific account. Added unit test to check balance on well known account.


rnistuk:
- 13 unit tests pass
- there are log messages that need to be removed (probably from a different set of commits):
```
Node created: 247.0.234.115
Node destroyed: 247.0.234.115
Node created: 249.189.30.45
Node destroyed: 249.189.30.45
Node created: 230.135.179.32
Node destroyed: 230.135.179.32
```
Let's get rid of those now.
- ethereum_api::token_balance is a bit large and there are blocks of code that can bet turned into functions to make it easier to read.
- Returning -1.0 is pretty arbitrary, would throwing a custom exception be better here?
- ethereum_token uses std::stringstream but does not include the sstream header. Let's not rely on accidents to make the project build.
- the format of ethereum_api does not follow Bluzelle coding guidelines.

rnistuk:
Accidentally closed the request. Reopening and  setting the base to develop.

rnistuk:
- please add a test for failure of the etherium_api

rnistuk:
CMakeLists.txt
- macOS doesn't need the pthread library to be linked. Wrap those directives in if(UNIX AND NOT APPLE) blocks

dmitrymukhin:
@rnistuk Cool, will fix shortly.

dmitrymukhin:
Reformatted
Fixed makefile
Split into smaller methods
added check for failure UT
Include missing header.

@rnistuk Could you take a look?

rnistuk:
- 
-

rnistuk:
- Node::generate_name, on_death, on_birth are only used by node, is there a reason they should not be private?
- In Node::generate_name you use boost::random::random_device, which is not available on all systems (http://www.boost.org/doc/libs/1_65_1/doc/html/boost/random/random_device.html). Use std::time(0) instead to initialize the mt19937
- I've been thinking about protected vs. private, and I think that to really enforce encapsulation we need to favour private, and then when required, move methods to protected only when  really necessary. We have no idea if we are going to extend the class at this time so why break encapsulation early?
- in EthereumToken why do we need address_ and decimal_points_ to be public? They should really be exposed with const getters as there is no reason to change them after instantiation.
- Node::generate_name should be unit tested by creating a node and verifying that Node::name returns a string who's format can be checked with regex. Ideally it should also be valid public IPv4, but checkin the format should be sufficient.
- since you can inject the listener, on_birth and on_death should also be unit tested.

rnistuk:
- 16 unit tests pass, good
- EtheriumAPI.h not following Whitsmith style
- EtheriumAPI connect_socket(), write_request, read_response, parse_response.. should be private unless the class is being extended, we should not solve future problems that may never exist.

dmitrymukhin:
@rnistuk I fixed EthereumApi.h any more comments?

rnistuk:
Great thanks.

PR #6, opened 2017-11-10
KEP-81: Merge KEP-81 to devel

scottburch:


bluzelled/inc/http/server/ServiceManager.h
```diff
@@ -0,0 +1,42 @@
+#ifndef KEPLER_SERVICEMANAGER_H
+#define KEPLER_SERVICEMANAGER_H
+
+#include "server.hpp"
+
+#include <boost/thread.hpp>
+
+namespace  http
```
dmitrymukhin: @rich This is the only file that uses namespaces. We should put them everywhere or remove from this one

bluzelled/inc/http/server/ServiceManager.h
```diff
@@ -0,0 +1,42 @@
...
+{
+    namespace  server
+    {
+        class ServiceManager
+        {
```
dmitrymukhin: @rich Brackets are in the wrong place :)

bluzelled/src/NodeUtilities.cpp
```diff
@@ -53,9 +52,47 @@ void remove_dead_nodes(Nodes *nodes)
...
 
-
 void reaper(Nodes *nodes)
 {
     join_dead_tasks(nodes);
     remove_dead_nodes(nodes);
 }
+
+void kill_nodes(const Nodes &nodes)
+{
```
dmitrymukhin: @rich Brackets -> "method_name() {"

bluzelled/src/NodeUtilities.cpp
```diff
@@ -53,9 +52,47 @@ void remove_dead_nodes(Nodes *nodes)
...
+    return new Node(l);
+}
+
+
+Nodes create_nodes(int number_of_nodes, std::shared_ptr<Listener> l)
+{
+    Nodes nodes;
+    for (int i = 0; i < number_of_nodes; ++i)
+        {
+        nodes.emplace_back(create_node(l));
```
dmitrymukhin: @rich AFAIK emplace_bake() method takes ctor parameters so it can be used as emplace_back(l). Create node method is redundant.

bluzelled/src/main.cpp
```diff
@@ -1,54 +1,39 @@
...
     boost::mutex::scoped_lock lock(*s_mutex);
     s_nodes.insert(s_nodes.end(), nodes.begin(), nodes.end());
 }
 
-void set_max_nodes(unsigned long max)
+void set_max_nodes
+        (
+                unsigned long max
+        )
 {
```
dmitrymukhin: @rich If I use reformat code feature from CLIon it always places curly brackets on the sam eline right after method name. In your case Node.h is like mine but min.cpp is different. I suggest running "Reformat Code" to set them properly.

bluzelled/src/main.cpp
```diff
@@ -63,33 +48,26 @@ unsigned long get_min_nodes()
...
+
         s_nodes = create_nodes(numTasks, g_listener);
         wait_for_all_nodes_to_start(s_nodes);
-        make_introductions(s_nodes);
 
         while (!s_nodes.empty())
             {
             reaper(&s_nodes);
-            int num_of_new_nodes = number_of_nodes_to_create(s_min_nodes, s_max_nodes, s_nodes.size());
+            unsigned long num_of_new_nodes = number_of_nodes_to_create(s_min_nodes, s_max_nodes, s_nodes.size());
```
dmitrymukhin: Not  a big deal, codingstandard mentions all parameters must be on its own line.

bluzelled/src/node_test.cpp
```diff
@@ -58,85 +49,168 @@ BOOST_AUTO_TEST_CASE( test_node_death )
...
+    BOOST_CHECK_EQUAL(true, all_nodes_alive(nodes));
+    BOOST_CHECK_EQUAL(k_node_count, nodes.size());
+    kill_and_join_all_nodes(nodes);
+}
+
+
+
+BOOST_AUTO_TEST_CASE(test_add_nodes)
+{
+
```
dmitrymukhin: If the test is not used then remove it, or finish.

bluzelled/src/services/service_tests.cpp
```diff
@@ -141,30 +146,65 @@ BOOST_FIXTURE_TEST_SUITE(websocket_services_tests, F)
...
         std::string actual(sut(cmd));
         BOOST_CHECK(string_to_ptree(accepted) == string_to_ptree(actual));
+
+        boost::asio::io_service ios;
+        tcp::endpoint endpoint;
+        Listener listener(ios, endpoint);
+        std::shared_ptr<Listener> shared_listener;
+
+
+        nodes.emplace_back(new Node(shared_listener));
```
dmitrymukhin: Again emplace_back(shared_listener) should do.

bluzelled/src/main.cpp
```diff
@@ -1,54 +1,39 @@
...
 static unsigned long s_min_nodes = 5;
 static Nodes s_nodes;
+
 static boost::mutex *s_mutex = nullptr;
 
 // List of nodes that were died and need to be reported to GUI.
 static Nodes s_removed_nodes;
 static boost::mutex s_removed_nodes_mutex;
-Nodes* get_removed_nodes() { return &s_removed_nodes; }
-boost::mutex& get_removed_nodes_mutex() { return s_removed_nodes_mutex; }
```
dmitrymukhin: s_removed_nodes, s_removed_nodes_mutex and corresponding methods are not used anymore. Seems like they are remnants of some merge.

bluzelled/src/main.cpp
```diff
@@ -1,54 +1,39 @@
...
-#include <boost/beast/version.hpp>
-#include <boost/asio/ip/tcp.hpp>
-#include <boost/asio/strand.hpp>
-#include <boost/config.hpp>
-#include <iostream>
-#include <string>
 
 static unsigned long s_max_nodes = 25;
 static unsigned long s_min_nodes = 5;
 static Nodes s_nodes;
```
dmitrymukhin: I a not big fan of defines for types. It's not clear what Nodes is. It looks like a container but is it vector, map, always have to go to definition to figure out.

bluzelled/inc/http/server/ServiceManager.h
rnistuk: Removed the brackets. and the namespace. I think the class bracket is in the right place...

rnistuk:
I think I've responded to all of the review messages. I'm going to close this pull request and create a new on as I have made a lot more changes to KEP-81

rnistuk:
Hi Dmitry,

Please have a look at my attempt to clean up main. I think there is still some work to do, but in other places. I've been able to move a lot of complexity out of main and into NodeManager. 

I've also made the changes you have already suggested.

Please review and make any suggestions you can.

Thanks,
  Rich.

PR #10, opened 2017-11-23
KEP-99: Merge KEP-99-node to node

scottburch:


dmitrymukhin:
@rnistuk ready for review/merge

rnistuk:
- You should move the std::cout's out of the CommandLineOptions class and provide a get_last_error method to tell the caller what happened if  CommandLineOptions::parse should ever return false. 
- CommandLineOptions::is_valid_ethereum_address has an empty catch block.
- CommandLineOptions::is_valid_ethereum_address(address) is not tested.
- Thanks for specifying the  "Service Name and Transport Protocol Port Number Registry" document, I was wondering where the port range game from. 


- Note that Mehdi will need to modify this class to add his command line parameters. In fact whenever we want to add a new command line option we'll have to modify the class. I wonder if CommandLineOptions should be more data driven. Maybe not as this is really a container for the boost program_options library.

PR #20, opened 2018-02-10
Added bash scripts to spawn nodes

dffffffff:
bash scripts for gnu screen and tmux and explanation for macOS


amastracci:
Thanks for the documentation pull request. We have restructured the repository and would like your help putting together a similar pull request for the new structure.

PR #22, opened 2018-04-27
Update README.md

amastracci:


README.md
```diff
@@ -1,91 +1,82 @@
...
-```
+If you want to deploy your swarm immediately you can use our docker-compose quickstart instructions:
 
-Feel free to use the "-j N" option with b2 to speed things up a bit -- you would need to determine how many CPU cores are on your machine. Basically, if you want to use multiple CPU cores, specify that number here in place of N. 
+Docker Swarm Deploy
 
-Now you will have the Boost 1.66.0 libraries installed in 
+1. Download sample [peerlist.json](https://gist.github.com/amastracci/b6144d32bd790cf1a93e3c498a9d46cc)
+2. Add your swarm machines' IP to the "host" field of each peer in peerlist.json
+3. Host your peerlist.json at a publically accessable URL
```
jyclam: Do we want to make a note that the file must be hosted in a non HTTPS and non redirecting URL?

README.md
```diff
@@ -1,91 +1,82 @@
...
 ```
-./bootstrap
-make
-sudo make install
+mkdir -p ~/mycmake
+curl -L http://cmake.org/files/v3.11/cmake-3.11.0-Darwin-x86_64.tar.gz | tar -xz -C ~/mycmake --strip-components=1
 ```
 
-Alternatively 
+Again this will result in a customer cmake install into ```~/mycmake/```
```
jyclam: custom?

README.md
amastracci: updated

README.md
amastracci: updated

README.md
```diff
@@ -112,12 +112,83 @@ Here are the steps to build the daemon and unit test application from the comman
...
+  "ethereum" : "0x006eae72077449caca91078ef78552c0cd9bce8f",
+  "uuid" : "60ba0788-9992-4cdb-b1f7-9f68eef52ab9",
+  "bootstrap_url" : "pastebin.com/raw/mbdezA9Z"
+}
+```
+
+Start your cluster from the daemon-build/output directory run:
+
+```
+swarm -c daemon-build/output -c bluzelle1.json
```
jyclam: Since the executable and the config files are in the same directory, it just needs to be 
```
./swarm -c bluzelle.json // you have bluzelle.json above, not bluzelle1.json
./swarm -c bluzelle2.json
./swarm -c bluzelle3.json
```

README.md
```diff
@@ -112,12 +112,83 @@ Here are the steps to build the daemon and unit test application from the comman
...
+ 
+ 
+// peers.json
+[
+  {"name": "peer1", "host": "127.0.0.1", "port": 50000, "uuid" : "60ba0788-9992-4cdb-b1f7-9f68eef52ab9"},
+  {"name": "peer2", "host": "127.0.0.1",  "port": 50001, "uuid" : "c7044c76-135b-452d-858a-f789d82c7eb7"},
+  {"name": "peer3", "host": "127.0.0.1",  "port": 50002, "uuid" : "3726ec5f-72b4-4ce6-9e60-f5c47f619a41"}
+]
+ 
+// bluzelle-bootstrap-url.json
```
jyclam: I would just not include this. This was only used for my tests to make sure that I could load a hosted peerlist. If you do decide to keep it, it should look like one of the configs above.

```
{
  "listener_address" : "127.0.0.1",
  "listener_port" : 50000,
  "ethereum" : "0x006eae72077449caca91078ef78552c0cd9bce8f",
  "bootstrap_url" : "http://pastebin.com/raw/eVfBShZJ",
  "uuid" : "60ba0788-9992-4cdb-b1f7-9f68eef52ab9",
  "debug_logging" : true,
  "log_to_stdout" : false
 
}
```

README.md
amastracci: yep.

README.md
amastracci: yep.

PR #24, opened 2018-05-09
KEP-295 Add threading support

ebruck:


rnistuk:
Project builds, all unit tests pass. 
OK - crud_tests.cpp: accepted response changed to a shared ptr to handle session::send_message changes
OK - crud.cpp: turned response into a shared pointer, very few changes to the rest of the module
OK - boost_asio_beast.hpp - Cool! We're using strands! The strand class looks good
OK - mock_ boost_asio_beast.hpp - io_context_base::make_unique_strand, strand_base::wrap and  strand_base::get_strand mocked
OK - mock_node_base.hpp - modified send_message to take a shared_ptr
OK - mock_session_base.hpp - needed to modify send_message to take a shared_ptr
OK - session_test.cpp - tests were modified to implement the strand multithreading functionality required for session
OK - node.cpp - two small changes in send_message, the msg parameter is now a shared pointer, and the session needs an io_context now, member of node.
OK - node.hpp - change to the send_message signature... and io_context already existed.
OK - node_base.hpp - same change to the send_message signature.
OK - session.cpp - constructor needs shared pointer io_context_base parameter. send_message signature modifed to take a shared_ptr message, and the websocket stuff is wrapped in a strand - thus, multithreaded... Nice
OK - session.hpp - session constructor and send_message signatures changed. Added strand object as a shared pointer. Memember send_msg removed due to not being used.
OK - session_base.cpp - same change to send_message
OK - raft_test.cpp - the raft tests that mocked send_message needed to be modified to change the request message to a shared pointer, or to dereference a shared pointer.
OK - main.cpp - oh noes!!! Main is getting a bit hairy... Yeah, during start up, start a number of threads (# set by std::thread::hardware_concurrency()) each having an io_context. Also there's a ping handler that needed the send_message changes to the response message.
OK - README.md - Changes to the docker instructions. *I have not tested this*. But the spelling looks good(?)

Yes, I'm cool with these changes.

PR #34, opened 2018-05-16
KEP-272 Implemented Value size limit

rnistuk:


raft/test/raft_test.cpp
```diff
@@ -21,11 +21,11 @@ using namespace ::testing;
 
 namespace
 {
-    const bzn::uuid_t TEST_NODE_UUID{"f0645cc2-476b-485d-b589-217be3ca87d5"};
+    const bzn::uuid_t k_TEST_NODE_UUID{"f0645cc2-476b-485d-b589-217be3ca87d5"};
```
ebruck: Not sure we need the k_ as all caps usually indicates const and our compiler will tell us so if we try to modify it.

storage/storage.cpp
```diff
@@ -50,6 +50,11 @@ storage::error_msg(storage_base::result error_id)
         case storage_base::result::exists:
             message = "exists";
             break;
+        case storage_base::result::value_too_large:
+            message = "value exceeds ";
```
ebruck: No need to break this into three lines:

mesage = "value exceeds " + std::to_string(MAX_VALUE_SIZE) + "  kb" 

should work fine.

If you stick with this then please put the code in a braces such as:

case ASDF:
{
    ...
    break;
}

storage/storage.hpp
```diff
@@ -40,7 +40,7 @@ namespace bzn
 
         storage_base::result load(const std::string& path) override;
 
-        std::string error_msg(storage_base::result error_id) override;
+        static std::string error_msg(storage_base::result error_id);// override;
```
ebruck: I would not pollute the public interface with this function. I'd put it into the unnamed namespace in the cpp file instead.
There is no need to make this a static and public.

storage/storage_base.hpp
```diff
@@ -64,7 +66,7 @@ namespace bzn
 
         virtual storage_base::result load(const std::string& path) = 0;
 
-        virtual std::string error_msg(storage_base::result error_id) = 0;
+        //virtual std::string error_msg(storage_base::result error_id) = 0;
```
ebruck: remove dead code

crud/test/crud_tests.cpp
```diff
@@ -126,7 +126,7 @@ class crud_test : public Test
 
 TEST_F(crud_test, test_that_follower_not_knowing_leader_fails_to_create)
 {
-    auto request = generate_create_request(user_uuid, "key0", k_TEST_VALUE);
+    auto request = generate_create_request(k_USER_UUID, "key0", k_TEST_VALUE);
```
ebruck: not a fan of k_

raft/test/raft_test.cpp
rnistuk: I will remove the k_ 's

storage/storage.cpp
rnistuk: I have combined the code into one line.

crud/test/crud_tests.cpp
rnistuk: Removed the k_ wart.

storage/storage.hpp
rnistuk: removed the error_msg function.

storage/test/storage_test.cpp
```diff
@@ -21,14 +21,15 @@
...
 
 using namespace ::testing;
 
 namespace
 {
-    const bzn::uuid_t user_uuid = "4bba2aeb-44fe-441e-bb6b-8817561eb716";
-    const std::string key = "bluzelle.json";
-    const std::string value = "ewogICJsaXN0ZW5lcl9hZGRyZXNzIiA6ICIxMjcuMC4wLjEiLAogICJsaXN0ZW5lcl9wb3J0IiA6"
+    const bzn::uuid_t USER_UUID = "4bba2aeb-44fe-441e-bb6b-8817561eb716";
+    const std::string k_KEY = "bluzelle.json";
```
ebruck: missed one

PR #41, opened 2018-05-29
KEP-277: RAFT persistence implementation

rnistuk:
Implements RAFT log state and log persistence. As a daemon commits entries, each entry is appended to a log file on local storage, at the same time the RAFT state variables are also persisted to a separate log file.

If the daemon is stopped and later restarted, the log entries are used to rehydrate the daemon to the point at which it was shut down.

In addition, the daemon storage is reloaded from the loaded log entries.

### DIFF
--- a/raft/log_entry.hpp
+++ b/raft/log_entry.hpp
@@ -1,0 +1,67 @@
+// Copyright (C) 2018 Bluzelle
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License, version 3,
+// as published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+
+#include <ostream>
+#include <boost/beast/core/detail/base64.hpp>
+#include <string>
+
+
+namespace bzn
+{
+    const std::string MSG_ERROR_ENCOUNTERED_INVALID_ENTRY_IN_LOG{"ENCOUNTERED_INVALID_ENTRY_IN_LOG"};
+
+
+    struct log_entry
+    {
+        friend std::ostream &operator<<(std::ostream& out, const log_entry& obj)
+        {
+            out << obj.log_index << " " << obj.term << " " << boost::beast::detail::base64_encode(obj.json_to_string(obj.msg)) << "\n";
+            return out;
+        }
+
+
+        friend std::istream &operator>>(std::istream& in, log_entry& obj)
+        {
+            std::string msg_string;
+            obj.log_index = std::numeric_limits<uint32_t>::max();;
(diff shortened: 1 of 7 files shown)


raft/log_entry.hpp
```diff
@@ -0,0 +1,58 @@
...
+            in >> msg_string;
+
+            Json::Reader reader;
+            reader.parse(boost::beast::detail::base64_decode(msg_string), obj.msg);
+            return in;
+        }
+
+
+        inline std::string
+        json_to_string(const bzn::message& msg) const
```
ebruck: Could you not just use toStyledString instead of the writer?

swarm/main.cpp
```diff
@@ -19,13 +19,12 @@
...
 #include <raft/raft.hpp>
 #include <storage/storage.hpp>
-
 #include <boost/log/expressions.hpp>
 #include <boost/log/support/date_time.hpp>
 #include <boost/log/utility/setup/common_attributes.hpp>
 #include <boost/log/utility/setup/file.hpp>
 #include <boost/program_options.hpp>
-
+#include <boost/filesystem.hpp>
```
ebruck: We should be using std::filesystem throughout our code.

raft/log_entry.hpp
```diff
@@ -0,0 +1,58 @@
+// Copyright (C) 2018 Bluzelle
```
ebruck: What happens when there is an error loading our state?

raft/log_entry.hpp
```diff
@@ -0,0 +1,58 @@
...
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
+#include <ostream>
+#include <boost/beast/core/detail/base64.hpp>
+
+namespace {
```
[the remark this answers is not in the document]
rnistuk: Removed, the empty anonymous namespace, and a couple of unused includes.

raft/log_entry.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+}
+
+
+namespace bzn {
+    struct log_entry {
+        friend std::ostream& operator << (std::ostream& out, const log_entry& obj)
+        {
+            std::string msg_string = obj.json_to_string(obj.msg);
```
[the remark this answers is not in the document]
rnistuk: Removed the variable, used obj.json_to_string directly.

raft/log_entry.hpp
rnistuk: toStyledString adds whitespace I'd rather not have in the log. I'd like to keep json_to_string for now, and hopefully remove it when we go to protobuf.

ebruck: Won't matter anyway... the payload is going to come from the client soon base64 encoded.

swarm/main.cpp
rnistuk: I agree, but until Apple get's it's poop together, I think we'll need boost. I will write a story in Jira about upgrading as soon as we can.

raft/log_entry.hpp
rnistuk: I have added some simple checks to the RAFT and log_entry rehydration.
Runtime exceptions with be thrown if
- the JSON messages in log entries cannot be parsed
- state values are not loaded
- after rehydrating the log entries, the state values do not agree with the values loaded from the entries log.

PR #42, opened 2018-05-31
KEP-321: Add easily matchable Daemon states logging

rnistuk:
Replaced ascii art with easily matchable state strings:

"RAFT State: Follower"
"RAFT State: Candidate"
"RAFT State: Leader"

There are no unit tests for this change.

### DIFF
--- a/raft/raft.cpp
+++ b/raft/raft.cpp
@@ -598,21 +598,21 @@ raft::update_raft_state(uint32_t term, bzn::raft_state state)
     switch (this->current_state)
     {
         case bzn::raft_state::leader:
-            LOG(info) << "RAFT State: Leader\n";
+            LOG(info) << "RAFT State: Leader";
 
             this->leader = this->uuid;
             break;
 
         case bzn::raft_state::candidate:
-            LOG(info) << "RAFT State: Candidate\n";
+            LOG(info) << "RAFT State: Candidate";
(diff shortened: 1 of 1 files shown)


raft/raft.cpp
```diff
@@ -598,35 +598,21 @@ raft::update_raft_state(uint32_t term, bzn::raft_state state)
...
-                         " ___                     _____     _ _                        \n"
-                         "|_ _|   __ _ _ __ ___   |  ___|__ | | | _____      _____ _ __ \n"
-                         " | |   / _` | '_ ` _ \\  | |_ / _ \\| | |/ _ \\ \\ /\\ / / _ \\ '__|\n"
-                         " | |  | (_| | | | | | | |  _| (_) | | | (_) \\ V  V /  __/ |   \n"
-                         "|___|  \\__,_|_| |_| |_| |_|  \\___/|_|_|\\___/ \\_/\\_/ \\___|_|  \n";
+            LOG(info) << "RAFT State: Follower\n";
             break;
 
         default:
+            LOG(info) << "RAFT State: Undefined(NB: This is bad!)\n";
```
ebruck: I don't think we need a "cute" remark here. ;-)

rnistuk: Have removed the new lines and the remark.

PR #43, opened 2018-06-01
KEP-307 Update CRUD to use protobuf messages

ebruck:


amastracci:
Please ensure that the appropriate user docs are updated as part of this commit.

PR #46, opened 2018-06-07
KEP-366 - Added functionality to find the last quorum type log_entry …

rnistuk:
The log entries are stored in a std::vector<bzn::log_entry> container. So to find the last quorum entry I wrote a method called

bzn::log_entry last_quorum()

which stupidly reverse iterates from the last log entry, log_entries.end(), to the first log entry, log_entries.begin(), looking for a log_entry that is not a bzn::log_entry_type::log_entry. This may become a problem when we have more log entry types, some of which are not involved in the quorum. So I wrote unit tests that will fail if that happens.

In later tasks I will refactor this method to make use of a state member in RAFT that will keep track of the last entered quorum log_entry, but that functionality has not been implemented yet.

### DIFF
--- a/raft/raft.cpp
+++ b/raft/raft.cpp
@@ -25,6 +25,7 @@ namespace
     const std::string MSG_ERROR_INVALID_LOG_ENTRY_FILE{"Invalid log entry file. Please Delete .state folder."};
     const std::string MSG_ERROR_EMPTY_LOG_ENTRY_FILE{"Empty log entry file. Please delete .state folder."};
     const std::string MSG_ERROR_INVALID_STATE_FILE{"Invalid state file. Please delete the .state folder."};
+    const std::string MSG_NO_PEERS_IN_LOG = "Unable to find peers in log entries.";
 
     const std::chrono::milliseconds DEFAULT_HEARTBEAT_TIMER_LEN{std::chrono::milliseconds(1000)};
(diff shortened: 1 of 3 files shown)


raft/raft.cpp
```diff
@@ -769,3 +765,21 @@ raft::perform_commit(uint32_t& commit_index, const bzn::log_entry& log_entry)
     this->save_state();
 }
 
+
+bzn::log_entry
+raft::last_quorum()
+{
+    // TODO: Speed this up by not doing a search, when a quorum entry is added, simply store the index. Perhaps only do the search if the index is wrong.
+    auto it = this->log_entries.end();
+    while(it != this->log_entries.begin())
```
isabelsavannah: can we just std::find here?

raft/raft.cpp
```diff
@@ -769,3 +765,21 @@ raft::perform_commit(uint32_t& commit_index, const bzn::log_entry& log_entry)
...
+    {
+        --it;
+        if  (it->entry_type != bzn::log_entry_type::log_entry)
+        {
+            break;
+        }
+    }
+    // assumes that the first entry in the log is always a quorum type.
+    return *it;
```
isabelsavannah: If that assumption is wrong, rather crash than return bad data

PR #50, opened 2018-06-14
KEP-290 Solidity to swarm without proxy

ebruck:


http/connection.cpp
```diff
@@ -0,0 +1,267 @@
...
+    std::stringstream post_data;
+    post_data << boost::beast::buffers(this->request.body().data());
+
+    // format request using protobuf...
+    database_msg request;
+    database_response response;
+    bzn::message msg;
+    msg["bzn-api"] = "crud";
+
+    if (path[REQUEST_PATH_IDX] == CREATE_REQ)
```
rnistuk: This if-else-tree is getting a bit large.

PR #51, opened 2018-06-15
Task/rnistuk/kep 365

rnistuk:
the first log_entry must be a single_quorum_definition created from the local or remote peers list.

Given a peers list, and a fresh start with no prior "state" or "dat" files, raft will create the dat and state files and append the peer list entry "single_quorum" to the empty log entry list and dat file.


rnistuk:
I do not want to merge this branch onto devel, I want to merge it onto KEP-283

PR #54, opened 2018-06-20
KEP-396 Create Python script to issue CRUD commands using protobuf

ebruck:


scripts/crud
```diff
@@ -0,0 +1,155 @@
...
+def send_request(ws, uuid, msg):
+    msg.db.header.db_uuid = uuid
+    msg.db.header.transaction_id = random.randint(1,sys.maxint)
+
+    req = {}
+    req["bzn-api"] = "database";
+    req["msg"] = base64.b64encode(msg.SerializeToString())
+
+    print msg
+    print "-" * 60 + '\n'
```
amastracci: Use "".format rather than '+' operator

scripts/crud
```diff
@@ -0,0 +1,155 @@
...
+    return msg
+
+
+if __name__ == "__main__":
+
+    parser = ArgumentParser(description="crud", formatter_class=RawDescriptionHelpFormatter)
+
+    parser.add_argument("-n", "--node", type=str, default=None, help="node's address (ex. 127.0.0.1:51010)")
+
+    parser.add_argument("-c", "--command", type=str, choices=["create", "read", "update", "delete", "has", "keys", "size"],
```
amastracci: use parser.add_subparsers to create another layer of commands. This will remove your if statement below.

scripts/crud
```diff
@@ -0,0 +1,215 @@
...
+        print("failed to read value from file: {}".format(e.strerror))
+        exit(1)
+
+
+
+
+    resp = args.func(args)
+
+    # save to a file...
+    if args.save:
```
ebruck: You've changed how this works. Save is only applicable to a read operation. It was intended to download a "value" as a file and not the entire response.

scripts/readme.md
```diff
@@ -0,0 +1,25 @@
...
+### Activate Virtual Environment
+    $ . ~/workspace/crud-app/bin/activate
+    
+### Install Dependencies
+
+    (crud-app)$ pip install websocket-client protobuf
+
+### Generate Protobuf Code
+    (crud-app)$ cd ../workspace/swarmdb/scripts
+    (crud-app)$ Protoc --python_out=../scripts ./bluzelle.proto ./database.proto
```
ebruck: protoc not Protoc

PR #59, opened 2018-06-22
Task/iscroggin/kep 368

isabelsavannah:
First commit is the changes from last time squashed, second is changes since then.


raft/raft.cpp
```diff
@@ -476,8 +477,56 @@ raft::handle_heartbeat_timeout(const boost::system::error_code& ec)
...
+
+    bzn::message json;
+    json["bzn-api"] = "audit";
+    json["audit-data"] = boost::beast::detail::base64_encode(msg.SerializeAsString());
+
+    auto json_ptr = std::make_shared<bzn::message>(json);
+
+    for (const auto& peer : this->peers) {
+        auto ep = boost::asio::ip::tcp::endpoint{boost::asio::ip::address_v4::from_string(peer.host), peer.port};
+        this->node->send_message(ep, json_ptr);
```
ebruck: I wouldn't bother with the temporary and construct and endpoint in the send_message call.

audit/audit.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+uint
+audit::error_count() const
+{
+    return this->recorded_errors.size();
+}
+
+void
+audit::start()
+{
```
ebruck: I like wrapping this in a lambda called by std::call_once

audit/audit.cpp
```diff
@@ -0,0 +1,107 @@
...
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+#include <audit/audit.hpp>
+#include <boost/beast/core/detail/base64.hpp>
+#include <boost/format.hpp>
+
+using namespace bzn;
+
+audit::audit(std::shared_ptr<bzn::node_base> node): node(node)
+{
```
ebruck: I'd format this as:

audit::audit(...)
    : node(node)
{
}
a bit easier to read

audit/audit.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+}
+
+const std::list<std::string>&
+audit::error_strings() const
+{
+    return this->recorded_errors;
+}
+
+uint
```
ebruck: please use size_t

audit/audit.cpp
ebruck: please also use std::move(node)

audit/audit.hpp
```diff
@@ -0,0 +1,48 @@
...
+    {
+    public:
+        audit(std::shared_ptr<bzn::node_base> node);
+
+        uint error_count() const override;
+
+        const std::list<std::string>& error_strings() const override;
+
+        void handle(const bzn::message& message, std::shared_ptr<bzn::session_base> session) override;
+        void handle_commit(const commit_notification&);
```
ebruck: shouldn't these also be in audit_base?

ebruck: My not need the abstract base class if we don't intend to mock this.

PR #60, opened 2018-06-23
Task/rnistuk/kep 367

rnistuk:
Leader Must Only Acknowledge Votes from Quorum

Given a single quorum or joint quorum, the leader must use the quorum to acknowledge only valid votes.

Isabel, please review my changes to make sure that I have not broken the functionality we worked on.

Ed, please review the code for correctness.

Thank You

### DIFF
--- a/raft/raft.cpp
+++ b/raft/raft.cpp
@@ -466,6 +466,27 @@ raft::handle_ws_raft_messages(const bzn::message& msg, std::shared_ptr<bzn::sess
         return;
     }
 
+
+    // TODO: Handle add_peer and remove_peer handlers here.
+    // - only the leader can do this
+    // - the current last quorum is not a joint quorum
+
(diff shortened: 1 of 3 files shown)


raft/raft.cpp
```diff
@@ -882,12 +889,138 @@ raft::last_quorum()
...
+    {
+        all_peers.emplace_back(log_entry.msg);
+    }
+    else
+    {
+        all_peers.emplace_back(log_entry.msg["old"]);
+        all_peers.emplace_back(log_entry.msg["new"]);
+    }
+
+    for(std::reference_wrapper<bzn::message> jpeers   : all_peers)
```
ebruck: auto should work here?

rnistuk: SWEET! Good call I'll make that change and see if it works.

PR #61, opened 2018-06-24
KEP-388, KEP-389: Audit module liveness checks

isabelsavannah:
Please pay particular attention to the audit_test class. My goal there is to avoid duplication by allowing derived tests to declare different expectations on mock timers without duplicating the code that creates and injects those timers. It works, but I'm not completely sure on the safety/best-practices-ness of it.


audit/audit.cpp
```diff
@@ -46,9 +48,74 @@ audit::start()
...
+    this->leader_alive_timer->expires_from_now(this->leader_timeout);
+    this->leader_alive_timer->async_wait(std::bind(&audit::handle_leader_alive_timeout, shared_from_this(), std::placeholders::_1));
+}
+
+void
+audit::handle_leader_alive_timeout(const boost::system::error_code& ec)
+{
+    if(ec)
+    {
+        LOG(debug) << "Leader alive timeout canceled " << ec;
```
ebruck: use: ec.message() instead of the error code.

audit/audit.hpp
```diff
@@ -38,13 +40,37 @@ namespace bzn
...
+
         std::map<uint64_t, bzn::uuid_t> recorded_leaders;
         std::map<uint64_t, std::string> recorded_commits;
 
         std::once_flag start_once;
+        std::mutex audit_lock;
+        std::unique_ptr<bzn::asio::steady_timer_base> leader_alive_timer;
+        std::unique_ptr<bzn::asio::steady_timer_base> leader_progress_timer;
+
+        std::chrono::milliseconds leader_timeout{std::chrono::milliseconds(30000)};
```
ebruck: Is this something we may want to configure?

audit/audit.hpp
```diff
@@ -38,13 +40,37 @@ namespace bzn
...
+
+        bzn::uuid_t last_leader;
+        uint64_t last_leader_commit_index;
```
ebruck: Set this to zero?

audit/audit.hpp
isabelsavannah: Yeah, but I was not including that in this commit to make it as simple as possible for the sake of Lovelace

PR #63, opened 2018-06-26
Task/rnistuk/kep 370

rnistuk:
Isabel, I'm concerned that the commit indexes are incorrect, can you make sure that I got them correct?

### DIFF
--- a/raft/test/raft_test.cpp
+++ b/raft/test/raft_test.cpp
@@ -1183,9 +1183,6 @@ namespace bzn
                     return true;
                 });
 
-
-
-
         // and away we go...
         raft->start();
 
@@ -1231,9 +1228,9 @@ namespace bzn
 
 
         EXPECT_CALL(*this->mock_node, send_message(_, _)).WillRepeatedly(Invoke(
-                [](const auto& /*session*/, const auto& msg)
+                [](const auto& /*session*/, const auto& /*msg*/)
                 {
-                    std::cout << msg->toStyledString();
+                    //std::cout << msg->toStyledString();
                 }));
 
 
@@ -1245,42 +1242,6 @@ namespace bzn
         raft->handle_request_append_entries_response(bzn::create_append_entries_response("uuid2", 1, true, 1), mock_session);
         raft->handle_request_append_entries_response(bzn::create_append_entries_response(TEST_NODE_UUID, 1, true, 1), mock_session);
         wh(boost::system::error_code());
-
-
-
-
-
-
-
-        std::cout << " test\n";
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
 }
 
 
@@ -1571,6 +1532,7 @@ namespace bzn
 
     TEST_F(raft_test, test_that_joint_quorum_is_converted_to_single_quorum_and_committed)
     {
+        // TODO make sure comit i=indexes are correct.
         auto mock_steady_timer = std::make_unique<NiceMock<bzn::asio::Mocksteady_timer_base>>();
 
         // intercept the timeout callback...
@@ -1606,11 +1568,6 @@ namespace bzn
                     return true;
                 });
 
-
-
-
-
-
         // and away we go...
         raft->start();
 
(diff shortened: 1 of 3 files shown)


raft/test/raft_test.cpp
```diff
@@ -575,8 +574,8 @@ namespace bzn
 
         resp.clear();
 
-        // send append entry with commit index of 1... and follower commits and updates its match index
```
isabelsavannah: Because these are all sent with the same prev_log_entry and prev_log_term (both 0), they should all overwrite each other.

raft/test/raft_test.cpp
```diff
@@ -575,8 +574,8 @@ namespace bzn
```
isabelsavannah: I think commitindex should be 1 on this first request

raft/test/raft_test.cpp
```diff
@@ -575,8 +574,8 @@ namespace bzn
 
         resp.clear();
 
-        // send append entry with commit index of 1... and follower commits and updates its match index
-        msg = bzn::create_append_entries_request(TEST_NODE_UUID, 2, 1, 0, 0, 2, entry);
+        // send append entry with commit index of 3... and follower commits and updates its match index
+        msg = bzn::create_append_entries_request(TEST_NODE_UUID, 2, 3, 0, 0, 2, entry);
```
isabelsavannah: I think this should have commit_index 2, prev_index 1

raft/test/raft_test.cpp
```diff
@@ -585,15 +584,15 @@ namespace bzn
         resp.clear();
 
         // send invalid entry. peer should return false and not move back past commit index
-        msg = bzn::create_append_entries_request(TEST_NODE_UUID, 2, 0, 0, 0, 0, entry);
+        msg = bzn::create_append_entries_request(TEST_NODE_UUID, 2, 1, 0, 0, 0, entry);
```
isabelsavannah: To test rejecting invalid entries, this should be at the next log index after the prev one, so commit_index 3, prev_index 2. It gets rejected because it's not from the current term?

raft/test/raft_test.cpp
```diff
@@ -575,8 +574,8 @@ namespace bzn
 
         resp.clear();
 
-        // send append entry with commit index of 1... and follower commits and updates its match index
-        msg = bzn::create_append_entries_request(TEST_NODE_UUID, 2, 1, 0, 0, 2, entry);
+        // send append entry with commit index of 3... and follower commits and updates its match index
+        msg = bzn::create_append_entries_request(TEST_NODE_UUID, 2, 3, 0, 0, 2, entry);
         mh(msg, this->mock_session);
         EXPECT_EQ(resp["data"]["matchIndex"].asUInt(), Json::UInt(1));
```
isabelsavannah: This will be 2

raft/test/raft_test.cpp
```diff
@@ -585,15 +584,15 @@ namespace bzn
         resp.clear();
 
         // send invalid entry. peer should return false and not move back past commit index
-        msg = bzn::create_append_entries_request(TEST_NODE_UUID, 2, 0, 0, 0, 0, entry);
+        msg = bzn::create_append_entries_request(TEST_NODE_UUID, 2, 1, 0, 0, 0, entry);
         mh(msg, mock_session);
         ASSERT_FALSE(resp["data"]["success"].asBool());
         EXPECT_EQ(resp["data"]["matchIndex"].asUInt(), Json::UInt(1));
```
isabelsavannah: This will be 2

raft/test/raft_test.cpp
```diff
@@ -585,15 +584,15 @@ namespace bzn
...
 
         resp.clear();
 
         // put back append entries - bad append entry
-        msg = bzn::create_append_entries_request(TEST_NODE_UUID, 2, 1, 0, 0, 2, entry);
+        msg = bzn::create_append_entries_request(TEST_NODE_UUID, 2, 2, 0, 0, 2, entry);
```
isabelsavannah: I'm not sure the intent of this message, but I _think_ it should have the same prev_index and commit_index as the last rejected message, and be rejected because its prev_term is wrong. To that end, the previous invalid message should have valid prev_term.

raft/test/raft_test.cpp
rnistuk: Thanks for your help Isabel.

PR #64, opened 2018-06-26
KEP-391: Audit module pushes to stats.d

isabelsavannah:


audit/audit.cpp
```diff
@@ -14,14 +14,21 @@
...
-audit::audit(std::shared_ptr<bzn::asio::io_context_base> io_context, std::shared_ptr<bzn::node_base> node)
+audit::audit(std::shared_ptr<bzn::asio::io_context_base> io_context
+        , std::shared_ptr<bzn::node_base> node
+        , bzn::optional<boost::asio::ip::udp::endpoint> monitor_endpoint)
+
         : node(node)
+        , io_context(io_context)
         , leader_alive_timer(io_context->make_unique_steady_timer())
         , leader_progress_timer(io_context->make_unique_steady_timer())
+        , monitor_endpoint(monitor_endpoint)
```
ebruck: std::move may work here

audit/audit.cpp
```diff
@@ -113,7 +131,23 @@ audit::report_error(const std::string& short_name, const std::string& descriptio
 {
     this->recorded_errors.push_back(description);
     LOG(fatal) << boost::format("[%1%]: %2%") % short_name % description;
-    // TODO: Push to stats.d goes here
+    this->send_to_monitor(str(boost::format("%1%:1|c") % short_name));
```
ebruck: It may be better to put the format in a "const std::string". 
In the future we may have more types such as guages, timing etc.

audit/audit.cpp
```diff
@@ -113,7 +131,23 @@ audit::report_error(const std::string& short_name, const std::string& descriptio
...
+    {
+        return;
+    }
+
+    LOG(debug) << boost::format("Sending stat '%1%' to monitor at %2%:%3%")
+                  % stat
+                  % this->monitor_endpoint->address().to_string()
+                  % this->monitor_endpoint->port();
+
+    this->socket->send_to(stat, *(this->monitor_endpoint));
```
ebruck: send_to is a blocking call. We try to avoid those as much as possible.
Please use async_send_to.

audit/audit.cpp
```diff
@@ -146,6 +180,7 @@ audit::handle_leader_status(const leader_status& leader_status)
     if(this->recorded_leaders.count(leader_status.term()) == 0)
     {
         LOG(info) << "audit recording that leader of term " << leader_status.term() << " is '" << leader_status.leader() << "'";
+        this->send_to_monitor(bzn::NEW_LEADER_METRIC_NAME+":1|c");
```
ebruck: const std::string?

audit/audit.cpp
```diff
@@ -214,6 +249,8 @@ audit::handle_commit(const commit_notification& commit)
 {
     std::lock_guard<std::mutex> lock(this->audit_lock);
 
+    this->send_to_monitor(bzn::COMMIT_METRIC_NAME + ":1|c");
```
ebruck: statsd format is duped in too many places

include/optional.hpp
```diff
@@ -0,0 +1,31 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+#pragma once
```
ebruck: hopefully this can go away with the next release of xcode

audit/audit.cpp
ebruck: You may need to wrap this in a strand.

isabelsavannah: The whole thing is under mutexes held by the top level message handlers

audit/audit_base.hpp
```diff
@@ -32,6 +32,8 @@ namespace bzn
     const std::string COMMIT_METRIC_NAME = "stats.commit_heard";
     const std::string NEW_LEADER_METRIC_NAME = "stats.new_leader_heard";
```
ebruck: We will want to better organize this as:

com.bluzelle.swarm.UUID.node.UUID.stats.commit_heard
com.bluzelle.swarm.UUID.node.UUID.stats.new_leader_heard

audit/audit.cpp
```diff
@@ -131,7 +131,7 @@ audit::report_error(const std::string& short_name, const std::string& descriptio
 {
     this->recorded_errors.push_back(description);
     LOG(fatal) << boost::format("[%1%]: %2%") % short_name % description;
-    this->send_to_monitor(str(boost::format("%1%:1|c") % short_name));
+    this->send_to_monitor(short_name + bzn::STATSD_COUNTER_FORMAT);
```
ebruck: short_name should really be called a "namespace"

audit/audit.cpp
```diff
@@ -147,7 +147,20 @@ audit::send_to_monitor(const std::string& stat)
                   % this->monitor_endpoint->address().to_string()
                   % this->monitor_endpoint->port();
 
-    this->socket->send_to(stat, *(this->monitor_endpoint));
+    this->socket->async_send_to(stat, *(this->monitor_endpoint), std::bind(&audit::handle_udp_send_callback,
+                                                                           shared_from_this(),
```
ebruck: We will need to wrap the callback in a strand.
There may be a chance that we will have a flurry of sends that may corrupt the data or trigger an asio internal assert if not protected.

audit/audit.cpp
```diff
@@ -147,7 +147,20 @@ audit::send_to_monitor(const std::string& stat)
                   % this->monitor_endpoint->address().to_string()
                   % this->monitor_endpoint->port();
 
-    this->socket->send_to(stat, *(this->monitor_endpoint));
+    this->socket->async_send_to(stat, *(this->monitor_endpoint), std::bind(&audit::handle_udp_send_callback,
```
ebruck: Stat is copied here?
Other ways would be to use a lambda and capture the data.

audit/audit.cpp
```diff
@@ -14,14 +14,23 @@
...
-audit::audit(std::shared_ptr<bzn::asio::io_context_base> io_context, std::shared_ptr<bzn::node_base> node)
-        : node(node)
+audit::audit(std::shared_ptr<bzn::asio::io_context_base> io_context
+        , std::shared_ptr<bzn::node_base> node
+        , bzn::optional<boost::asio::ip::udp::endpoint> monitor_endpoint
+        , bzn::uuid_t uuid)
+
+        : uuid(std::move(uuid))
+        , node(node)
+        , io_context(io_context)
```
ebruck: std::move(io_context)

audit/audit.cpp
```diff
@@ -228,6 +278,12 @@ audit::handle_commit(const commit_notification& commit)
...
                               % commit.operation());
-        this->report_error("commitconflict", err);
+        this->report_error(bzn::COMMIT_CONFLICT_METRIC_NAME, err);
     }
 }
+
+std::string
+audit::build_statsd_prefix()
+{
+    return "com.bluzelle.swarm.singleton.node." + this->uuid;
```
ebruck: Could this not be built at construction time instead of calling this over and over?

PR #66, opened 2018-06-26
KEP-283 Dynamic Swarm Peer Membership

rnistuk:
UPDATE
Isabel and I removed the logging functionality out of RAFT and into its own class to reduce confusion
regarding when something is logged as opposed to when it is committed (mainly my confusion).

I also fixed and intermittent bug w.r.t. writing the initial log entry to a file in the .state directory when the .state directory does not exist. 
END UPDATE

Here is the Dynamic Swarm Peer Membership functionality. Please have a quick look. Each sub task has been reviewed.

Note that this pull request also includes the change that increases the output of append entry log messages from 60 to 1024 characters.

### DIFF
--- a/bootstrap/bootstrap_peers.cpp
+++ b/bootstrap/bootstrap_peers.cpp
@@ -56,24 +56,10 @@ bootstrap_peers::get_peers() const
 }
 
 
-bool 
-bootstrap_peers::ingest_json(std::istream& peers)
+size_t
+bootstrap_peers::initialize_peer_list(const Json::Value& root, bzn::peers_list_t& /*peer_addresses*/)
 {
-    size_t addresses_before = this->peer_addresses.size();
     size_t valid_addresses_read = 0;
-
-    Json::Value root;
-
-    try
-    {
-        peers >> root;
-    }
-    catch (const std::exception& e)
-    {
-        LOG(error) << "Failed to parse peer JSON (" << e.what() << ")";
-        return false;
-    }
-
     // Expect the read json to be an array of peer objects
     for (const auto& peer : root)
     {
@@ -123,6 +109,27 @@ bootstrap_peers::ingest_json(std::istream& peers)
 
         valid_addresses_read++;
     }
+    return valid_addresses_read;
+}
+
+
+bool
+bootstrap_peers::ingest_json(std::istream& peers)
+{
+    size_t addresses_before = this->peer_addresses.size();
+    Json::Value root;
+
+    try
+    {
+        peers >> root;
+    }
+    catch (const std::exception& e)
+    {
+        LOG(error) << "Failed to parse peer JSON (" << e.what() << ")";
+        return false;
+    }
+
+    size_t valid_addresses_read = this->initialize_peer_list(root, this->peer_addresses);
 
     size_t new_addresses = this->peer_addresses.size() - addresses_before;
     size_t duplicate_addresses = new_addresses - valid_addresses_read;
--- a/crud/crud.cpp
+++ b/crud/crud.cpp
@@ -54,6 +54,13 @@ crud::start()
                 {
                     bzn_msg msg;
 
+                    // This is to avoid trying to process encoded protobuf messages where we just want to commit this
+                    // index, but not alter storage at all.
+                    if(!ws_msg["msg"].isString())
+                    {
+                        return true;
+                    }
+
                     if (msg.ParseFromString(boost::beast::detail::base64_decode(ws_msg["msg"].asString())))
                     {
                         if (msg.msg_case() == bzn_msg::kDb)
@@ -105,11 +112,9 @@ crud::handle_create(const bzn::message& msg, const database_msg& request, databa
 
     if (this->raft->get_state() == bzn::raft_state::leader)
     {
-        this->raft->append_log(msg);
+        this->raft->append_log(msg, bzn::log_entry_type::database);
         return;
     }
-
-    this->set_leader_info(response);
 }
 
 
@@ -149,7 +154,7 @@ crud::handle_update(const bzn::message& msg, const database_msg& request, databa
 
     if (this->raft->get_state() == bzn::raft_state::leader)
     {
-        this->raft->append_log(msg);
+        this->raft->append_log(msg, bzn::log_entry_type::database);
         return;
     }
 
@@ -168,7 +173,7 @@ crud::handle_delete(const bzn::message& msg, const database_msg& request, databa
 
     if (this->storage->has(request.header().db_uuid(), request.delete_().key()))
     {
-        this->raft->append_log(msg);
+        this->raft->append_log(msg, bzn::log_entry_type::database);
         return;
     }
 
--- a/include/bluzelle.hpp
+++ b/include/bluzelle.hpp
@@ -48,3 +48,6 @@ namespace bzn::utils
 
 // logging
 #define LOG(x) BOOST_LOG_TRIVIAL(x) << "(" << bzn::utils::basename(__FILE__) << ":"  << __LINE__ << ") - "
+
+// This limits the number of characters that are displayed in "LOG(x) <<"  messages.
+const uint16_t MAX_MESSAGE_SIZE = 1024;
--- a/raft/raft.cpp
+++ b/raft/raft.cpp
@@ -27,7 +27,6 @@ namespace
     const std::string MSG_ERROR_EMPTY_LOG_ENTRY_FILE{"Empty log entry file. Please delete .state folder."};
     const std::string MSG_ERROR_INVALID_STATE_FILE{"Invalid state file. Please delete the .state folder."};
     const std::string MSG_NO_PEERS_IN_LOG{"Unable to find peers in log entries."};
-
     const std::chrono::milliseconds DEFAULT_HEARTBEAT_TIMER_LEN{std::chrono::milliseconds(1000)};
     const std::chrono::milliseconds  DEFAULT_ELECTION_TIMER_LEN{std::chrono::milliseconds(5000)};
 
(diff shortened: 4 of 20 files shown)


node/session.cpp
```diff
@@ -184,7 +184,7 @@ session::start_idle_timeout()
 {
     this->idle_timer->cancel();
 
-    LOG(debug) << "resetting " << this->ws_idle_timeout.count() << "ms idle timer";
+    //LOG(debug) << "resetting " << this->ws_idle_timeout.count() << "ms idle timer";
```
ebruck: do not remove this

raft/raft.hpp
```diff
@@ -138,6 +138,8 @@ namespace bzn
         void notify_leader_status();
         void notify_commit(size_t log_index, const std::string& operation);
 
+        //uint32_t last_log_index();
```
ebruck: nuke dead code

raft/raft.cpp
```diff
@@ -304,49 +304,54 @@ raft::handle_ws_append_entries(const bzn::message& msg, std::shared_ptr<bzn::ses
...
-                    LOG(debug) << "at commit index: " << this->commit_index;
-                }
-            }
+            LOG(debug) << "Rejecting AppendEntries because I do not agree with the previous index";
         }
+        success = false;
     }
 
-    auto resp_msg = std::make_shared<bzn::message>(bzn::create_append_entries_response(this->uuid, this->current_term, success, this->last_log_index));
+    auto match_index = std::min(this->log_entries.size(), (size_t) msg_index+1);
```
ebruck: don't use auto for POD types

node/session.cpp
rnistuk: Uncommented the line.

raft/raft.hpp
rnistuk: Terminated with extreme prejudice.

raft/raft.cpp
rnistuk: Fixed.

crud/crud.cpp
```diff
@@ -54,6 +54,12 @@ crud::start()
                 {
                     bzn_msg msg;
 
+                    // TODO: This is a hack to stop a crash
```
ebruck: Your comment is misleading. It should read:

This is to avoid trying to process encoded protobuf messages where we just want to commit this index, but not alter storage at all.

rnistuk: Cool, making the change.

rnistuk:
OK, one more time.

Isabel and I ripped logging out of RAFT, as it seemed to be confused with consensus. We do not wait for consensus to put an entry into the log..

I hope we fixed that.

include/bluzelle.hpp
```diff
@@ -48,3 +48,4 @@ namespace bzn::utils
 
 // logging
 #define LOG(x) BOOST_LOG_TRIVIAL(x) << "(" << bzn::utils::basename(__FILE__) << ":"  << __LINE__ << ") - "
+const uint16_t MAX_MESSAGE_SIZE = 1024;
```
ebruck: can we add a space and comment describing what this is for?

swarm/main.cpp
```diff
@@ -64,6 +64,32 @@ init_logging()
 }
 
 
+void init_peers(bzn::bootstrap_peers& peers, const std::string& peers_file, const std::string& peers_url)
```
ebruck: return on a separate line

raft/raft_log.cpp
```diff
@@ -0,0 +1,165 @@
...
+#include "raft_log.hpp"
+
+#include <fstream>
+#include <boost/filesystem/operations.hpp>
+#include <iostream>
+
+
+namespace bzn
+{
+    raft_log::raft_log(const std::string& log_path) : entries_log_path(log_path)
```
ebruck: init on a newline

raft/raft_log.cpp
```diff
@@ -0,0 +1,165 @@
...
+    {
+        if(boost::filesystem::exists(this->entries_log_path))
+        {
+            std::ifstream is(this->entries_log_path, std::ios::in | std::ios::binary);
+            bzn::log_entry log_entry;
+            while (is >> log_entry)
+            {
+                this->log_entries.emplace_back(log_entry);
+            }
+            is.close();
```
ebruck: destructor will close

raft/raft_log.cpp
```diff
@@ -0,0 +1,165 @@
...
+        return this->log_entries.at(i);
+    }
+
+
+    const bzn::log_entry&
+    raft_log::last_quorum_entry() const
+    {
+        // TODO: Speed this up by not doing a search, when a quorum entry is added, simply store the index. Perhaps only do the search if the index is wrong.
+        auto result = std::find_if(
+                this->log_entries.crbegin(),
```
ebruck: too many indents

raft/raft_log.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+    void
+    raft_log::append_log_disk(const bzn::log_entry& log_entry)
+    {
+        if (!this->log_entry_out_stream.is_open())
+        {
+            boost::filesystem::path path{this->entries_log_path};
+            if(!boost::filesystem::exists(path.parent_path()))
+            {
+                boost::filesystem::create_directories(path.parent_path());
```
ebruck: should we not check to see if the path is created?

raft/raft_log.cpp
```diff
@@ -0,0 +1,165 @@
...
+        {
+            throw std::runtime_error(MSG_TRYING_TO_INSERT_INVALID_ENTRY);
+        }
+    }
+
+
+    bool
+    raft_log::entry_accepted(size_t previous_index, size_t previous_term) const
+    {
+        if(previous_index < this->log_entries.size()-1)
```
ebruck: do we need to guard against a size of 0?

raft/raft_log.cpp
```diff
@@ -0,0 +1,165 @@
...
+            return false;
+        }
+        return previous_term == this->log_entries[previous_index].term;
+    }
+
+
+    void
+    raft_log::rewrite_log()
+    {
+        boost::filesystem::path path{this->entries_log_path};
```
ebruck: Could path be a member variable that's not constructed over and over?

raft/raft.cpp
```diff
@@ -38,37 +37,38 @@ namespace
 using namespace bzn;
 
 
-raft::raft(std::shared_ptr<bzn::asio::io_context_base> io_context, std::shared_ptr<bzn::node_base> node, const bzn::peers_list_t& peers, bzn::uuid_t uuid, const std::string state_dir)
+raft::raft(
+        std::shared_ptr<bzn::asio::io_context_base> io_context
+        , std::shared_ptr<bzn::node_base> node
+        , const bzn::peers_list_t& peers
+        , bzn::uuid_t uuid, const std::string state_dir)
```
ebruck: state dir on newline

raft/raft.cpp
```diff
@@ -38,37 +37,38 @@ namespace
...
     }
+    this->setup_peer_tracking(peers);
+    this->get_raft_timeout_scale();
 
-    // setup peer tracking...
-    for (const auto& peer : this->peers)
-    {
-        this->peer_match_index[peer.uuid] = 0;
-    }
+    const std::string log_path("./.state/" + this->uuid + ".dat");
```
ebruck: use new state_dir variable

raft/raft.cpp
```diff
@@ -372,12 +366,136 @@ raft::handle_ws_append_entries(const bzn::message& msg, std::shared_ptr<bzn::ses
 }
 
 
+bzn::message
+raft::create_joint_quorum_by_adding_peer(const bzn::message& last_quorum_message, const bzn::message& new_peer)
```
ebruck: I'd move these three functions to raft base with the others for consistency

raft/raft.cpp
```diff
@@ -617,39 +727,38 @@ raft::handle_request_append_entries_response(const bzn::message& msg, std::share
...
 
     return true;
 }
 
 
+bool
+raft::append_log(const bzn::message& msg, const bzn::log_entry_type entry_type)
+{
+    std::lock_guard<std::mutex> lock(this->raft_lock);
+    return this->append_log_unsafe(msg, entry_type);
```
ebruck: you are locking to I guess unsafe is now safe?

raft/raft.cpp
```diff
@@ -792,49 +884,229 @@ raft::load_state()
...
-    while (is >> log_entry)
+    this->notify_commit(commit_index, log_entry.json_to_string(log_entry.msg));
+    this->commit_handler(log_entry.msg);
+    commit_index++;
+    this->save_state();
+
+    if(this->get_state() == bzn::raft_state::leader && log_entry.entry_type == bzn::log_entry_type::joint_quorum)
     {
-        this->log_entries.emplace_back(log_entry);
+        this->append_log_unsafe(
```
ebruck: you lock elsewhere. why not here?

raft/raft.cpp
```diff
@@ -792,49 +884,229 @@ raft::load_state()
...
+{
+    // TODO: cache this method when the active quorum is updated
+
+    auto extract_uuid = [](const auto& node_json) -> bzn::uuid_t
+        {
+            return node_json["uuid"].asString();
+        };
+
+    bzn::log_entry log_entry = this->raft_log->last_quorum_entry();
+    switch(log_entry.entry_type) {
```
ebruck: no k&r

raft/raft.cpp
```diff
@@ -792,49 +884,229 @@ raft::load_state()
...
+    }
+    return result;
+}
+
+
+bool
+raft::in_quorum(const bzn::uuid_t& uuid)
+{
+    const auto quorums = this->get_active_quorum();
+    for(const auto q : quorums)
```
ebruck: const auto& q?

ebruck: or:
for (const auto& q : this->get_active_quarun())
...

swarm/main.cpp
```diff
@@ -64,6 +64,32 @@ init_logging()
...
+
+    if (!peers_url.empty())
+    {
+        peers.fetch_peers_from_url(peers_url);
+    }
+
+    if (peers.get_peers().empty())
+    {
+        LOG(error) << "Failed to find any bootstrap peers";
+        std::exit(5); // EXC_SOFTWARE
```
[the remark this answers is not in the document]
rnistuk: Fixed, see above. I chose 5 because it the magic number that says there was a software error.

bootstrap/bootstrap_peers.cpp
```diff
@@ -123,6 +109,29 @@ bootstrap_peers::ingest_json(std::istream& peers)
...
+    {
+        peers >> root;
+    }
+    catch (const std::exception& e)
+    {
+        LOG(error) << "Failed to parse peer JSON (" << e.what() << ")";
+        return false;
+    }
+
+    valid_addresses_read = this->initialize_peer_list(root, this->peer_addresses);
```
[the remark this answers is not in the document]
rnistuk: Done.

swarm/main.cpp
rnistuk: shit... OK, I missed that one. Fixed

raft/raft_log.cpp
rnistuk: fooook.  Fixed.

raft/raft_log.cpp
rnistuk: Done. Removed on of suspenders or belt...

raft/raft_log.cpp
rnistuk: Reformatted. Done

raft/raft_log.cpp
rnistuk: Done, I chose to use the non-throwing version of create_directories in this way:

boost::system::error_code ec;
if(!boost::filesystem::create_directories(path.parent_path(), ec))
{
    LOG(error) << "Unable to create path " << path.parent_path() << " with error code " << ec <<".";
    throw std::runtime_error(MSG_EXITING_DUE_TO_LOG_PATH_CREATION_FAILURE);
}

so that I didn't need to catch and re throw an exception

raft/raft_log.cpp
rnistuk: Technically, no,  log_entries will always have at least one entry, that being the first single quorum.

raft/raft_log.cpp
rnistuk: Good point. Turns out that boost::filesystem::remove will take a std::string.. Fixed.

raft/raft.cpp
rnistuk: Done.

raft/raft.cpp
rnistuk: Done.

raft/raft.cpp
rnistuk: We pulled the code that is in append_log_unsafe from this function, so we could reuse it in it's unsafe form without blocking.

raft/raft.cpp
rnistuk: This is where we would get thread locking if we used the "safe" version as we are already within the scope of 
     std::lock_guard<std::mutex> lock(this->raft_lock)
from the top of handle_ws_raft_messages.

raft/raft.cpp
rnistuk: arg.... fixing.

raft/raft.cpp
rnistuk: Cleaned up.

bootstrap/bootstrap_peers.cpp
```diff
@@ -123,6 +109,27 @@ bootstrap_peers::ingest_json(std::istream& peers)
 
         valid_addresses_read++;
     }
+    return valid_addresses_read;
+}
+
+
+bool
+bootstrap_peers::ingest_json(std::istream& peers)
```
ebruck: I really dislike calling this "injest"

rnistuk: heh, It's totally in jest!

rnistuk: At least it's not called "moist"

PR #67, opened 2018-06-27
KEP-416 Websocket assert fires for read & close operations

ebruck:


node/session.cpp
```diff
@@ -35,19 +35,7 @@ session::session(std::shared_ptr<bzn::asio::io_context_base> io_context, std::sh
...
-        boost::system::error_code ec;
-
-        this->websocket->get_websocket().close(boost::beast::websocket::close_code::normal, ec);
-
-        if (ec)
-        {
-           LOG(debug) << "failed to close websocket: " << ec.message();
-        }
-    }
+    //LOG(debug) << "shutdown";
```
amastracci: uncomment or remove

node/session.cpp
```diff
@@ -167,14 +156,13 @@ session::close()
...
+            [self = shared_from_this()](auto ec)
+            {
+                if (ec)
                 {
-                    if (ec)
-                    {
-                        LOG(error) << "failed to close websocket: " << ec.message();
-                    }
-                }));
+                    LOG(error) << "failed to close websocket: " << ec.message();
```
amastracci: Is this fatal? Should *never* happen, right?

node/session.cpp
ebruck: more for debugging but  I will nuke it

node/session.cpp
ebruck: nope... it's probably already closed

PR #70, opened 2018-06-29
KEP-390: Audit module discards old data

isabelsavannah:


options/options.cpp
```diff
@@ -258,6 +259,19 @@ options::get_ws_idle_timeout() const
...
+size_t
+options::get_audit_mem_size() const
+{
+    if (this->config_data.isMember(AUDIT_MEM_SIZE_KEY))
+    {
+        return this->config_data[AUDIT_MEM_SIZE_KEY].asUInt();
+    }
+    else
+    {
+        return 10000;
```
ebruck: Can we define this and others in a unnamed namespace?

PR #76, opened 2018-07-10
KEP-112 : KEP-450 and KEP-452 Setting and Display Storage Limits

rnistuk:
The functionality to allow us to set the maximum storage setting in the config file, and display the maximum storage value and current storage size (the size in bytes of the .dat file).

### DIFF
--- a/options/options_base.hpp
+++ b/options/options_base.hpp
@@ -20,12 +20,16 @@
 #include <include/optional.hpp>
 #include <include/boost_asio_beast.hpp>
 #include <string>
+#include <map>
 
 namespace bzn
 {
+
     class options_base
     {
     public:
+        // Suffixes for the max size parser.
+        static const std::map<char, size_t> BYTE_SUFFIXES;//{{'B', 1}, {'K', 1024}, {'M', 1048576}, {'G', 1073741824}, {'T', 1099511627776}};
 
         virtual ~options_base() = default;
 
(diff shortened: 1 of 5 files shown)


options/options_base.hpp
```diff
@@ -20,9 +20,14 @@
 #include <include/optional.hpp>
 #include <include/boost_asio_beast.hpp>
 #include <string>
+#include <map>
 
 namespace bzn
 {
+    // Suffixes for the max size parser.
+    const std::map<char, size_t> BYTE_SUFFIXES{{'B', 1}, {'K', 1024}, {'M', 1048576}, {'G', 1073741824}, {'T', 1099511627776}};
```
ebruck: I would think about moving this into options_base and making it public.

swarm/main.cpp
```diff
@@ -135,7 +147,8 @@ print_banner(const bzn::options& options, double eth_balance)
        << "   Ethereum Address ID: " << options.get_ethererum_address()  << "\n"
        << "      Local IP Address: " << options.get_listener().address().to_string() << "\n"
        << "               On port: " << options.get_listener().port()    << "\n"
-       << "         Token Balance: " << eth_balance << " ETH\n"
+       << "         Token Balance: " << eth_balance << " ETH" << "\n"<< "     Maximimum Storage: " << options.get_max_storage() << " Bytes" << "\n"
```
ebruck: put Maximum Storage code on a newline

options/options_base.hpp
rnistuk: Cool. Moved BYTE_SUFFIXES into options_base, and made it static. Also propagated the change through the code.

swarm/main.cpp
rnistuk: Done.

PR #77, opened 2018-07-13
KEP-112: If RAFT exceeds the max storage set in the options file, it …

rnistuk:
…will throw a run time exception which shold result in an exit

The sketchiest change is in
  raft::append_log_unsafe
where if raft ever exceeds it's maximum storage limit it will simply call raise(SIGINT) after throwing up an error message on the console. That occurs at the moment an append entry happens when the max storage has already been exceed.

There are also a few refactorings:
- pulled out add_peer and remove_peer remove functionality into their own methods
- after some pain with double '/' in the raft::entries_log_path() and raft::state_path() methods, changed them to use boos::filesystem
-

### DIFF
--- a/raft/raft.cpp
+++ b/raft/raft.cpp
@@ -450,6 +450,14 @@ raft::handle_add_peer(std::shared_ptr<bzn::session_base> session, const bzn::mes
         return;
     }
 
+    if(!peer.isMember("uuid") || peer["uuid"].asString().empty())
+    {
+        bzn::message response;
+        response["error"] = ERROR_INVALID_UUID;
+        session->send_message(std::make_shared<bzn::message>(response), true);
+        return;
+    }
+
     this->peer_match_index[peer["uuid"].asString()] = 1;
     this->append_log_unsafe(this->create_joint_quorum_by_adding_peer(last_quorum_entry.msg, peer),
                             bzn::log_entry_type::joint_quorum);
@@ -478,10 +486,10 @@ raft::handle_remove_peer(std::shared_ptr<bzn::session_base> session, const std::
     }
 
     const auto &peers = last_quorum_entry.msg["msg"]["peers"];
-    const auto &same_peer = std::find_if(
-            peers.begin(), peers.end(), [&](const auto &p)
+    const auto &same_peer = std::find_if(peers.begin(), peers.end(),
+            [&](const auto &peer)
             {
-                return p["uuid"].asString() == uuid;
+                return peer["uuid"].asString() == uuid;
             });
     if (same_peer == peers.end())
     {
@@ -511,6 +519,14 @@ raft::handle_ws_raft_messages(const bzn::message& msg, std::shared_ptr<bzn::sess
     }
     else if(msg["cmd"].asString() == "remove_peer" )
     {
+        if (!msg.isMember("data") || !msg["data"].isMember("uuid") || msg["data"]["uuid"].asString().empty())
+        {
+            bzn::message response;
+            response["error"] = ERROR_INVALID_UUID;
+            session->send_message(std::make_shared<bzn::message>(response), true);
+            return;
+        }
+
         this->handle_remove_peer(session, msg["data"]["uuid"].asString());
         return;
     }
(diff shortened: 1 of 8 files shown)


raft/raft.cpp
```diff
@@ -407,7 +416,85 @@ raft::create_joint_quorum_by_removing_peer(const bzn::message& last_quorum_messa
...
+                                         });
+    if (same_peer != last_quorum_entry.msg["msg"]["peers"].end())
+    {
+        bzn::message response;
+        response["error"] = ERROR_PEER_ALREADY_EXISTS;
+        session->send_message(std::make_shared<bzn::message>(response), true);
+        return;
+    }
 
+    this->peer_match_index[peer["uuid"].asString()] = 1;
```
ebruck: should we check that uuid is there otherwise we'll get an empty string back?

raft/raft.cpp
```diff
@@ -407,7 +416,85 @@ raft::create_joint_quorum_by_removing_peer(const bzn::message& last_quorum_messa
...
+    {
+        bzn::message response;
+        response["error"] = MSG_ERROR_CURRENT_QUORUM_IS_JOINT;
+        session->send_message(std::make_shared<bzn::message>(response), true);
+        return;
+    }
+
+    const auto &peers = last_quorum_entry.msg["msg"]["peers"];
+    const auto &same_peer = std::find_if(
+            peers.begin(), peers.end(), [&](const auto &p)
```
ebruck: I like putting the first to params on the same line as find_if and the lamda on the next indented.  You're call.

raft/raft.cpp
```diff
@@ -407,7 +416,85 @@ raft::create_joint_quorum_by_removing_peer(const bzn::message& last_quorum_messa
...
+            {
+                return p["uuid"].asString() == uuid;
```
ebruck: p?

raft/raft.cpp
```diff
@@ -60,7 +62,14 @@ raft::raft(
...
                                : this->create_state_files(log_path, peers);
 
-    this->raft_log = std::make_shared<bzn::raft_log>(log_path);
+    this->raft_log = std::make_shared<bzn::raft_log>(log_path, maximum_raft_storage);
+
+    // KEP-112 Bail if the raft storage exceeds the max storage
+    if(this->raft_log->maximum_storage_exceeded())
+    {
+        LOG(error) << MSG_ERROR_MAXIMUM_STORAGE_EXCEEDED;
+        throw std::runtime_error(MSG_ERROR_MAXIMUM_STORAGE_EXCEEDED);
```
ebruck: should we use just one way to exit?

raft/raft_log.hpp
```diff
@@ -45,17 +49,27 @@ namespace bzn
...
+            return this->maximum_storage < this->total_memory_used;
+        }
+
     private:
         void append_log_disk(const bzn::log_entry& log_entry);
         void rewrite_log();
 
         std::vector<log_entry> log_entries;
+        size_t                  total_memory_used = 0;
+        size_t                  maximum_storage;
```
ebruck: this should be a const and initialized at construction?

raft/raft.cpp
rnistuk: I have added a catch that will check if "uuid" does not exist, or the value is empty, then add peer will fail with a message.

Also created unit tests to exercise the bad add and remove paths... 

But, for UUID only. I will sneak in the tests for all the other parameters in future refactorings.

raft/raft.cpp
rnistuk: Cool. Done.

raft/raft_log.hpp
rnistuk: Yes. Changes maximum_storage to a const. It was already being initialized at construction.

raft/raft.cpp
rnistuk: I want to leave this one as a throw simply because I was able to unit test the functionality easily...

I'd like to push back on this one, but, it's easy enough to replace the throw with raise(SIGINT)...

PR #80, opened 2018-07-17
KEP-469: Fixed an out of bounds issue w.r.t. this->raft_log on line 7…

rnistuk:
…54, also a similar < to >= issue that resulted in the wrong debug message being presented on line 340

This fixes the crash when a follower peer is removed from a two node swarm.


rnistuk:
This pull request was submitted with failing unit tests. Closing it and starting again.

PR #82, opened 2018-07-19
Task/rnistuk/kep 488

rnistuk:
Here's the unit test that shows that 2 yes votes out of 4 will not form a consensus. I set up a raft instant with 3 peers, and have that peer become a candidate and tally votes.


rnistuk:
Whups, wrong branch to merge to..

PR #84, opened 2018-07-19
KEP-281: PBFT, Minimal normal case operation (kep-309, 334, 335, 336,

isabelsavannah:
337, 348, 379)


.gitignore
```diff
@@ -1,8 +1,10 @@
 .idea/
+.state/
```
isabelsavannah: making sure I can comment

pbft/test/CMakeLists.txt
```diff
@@ -0,0 +1,6 @@
+set(test_srcs pbft_test.cpp pbft_operation_test.cpp)
+set(test_libs pbft protobuf bootstrap)
+
+add_gmock_test(pbft_test)
```
ebruck: remove _test from these. It's now added for you.

ebruck: Your tests also have a lot of gmock warnings. Please have a look to clean that up.

pbft/pbft.cpp
```diff
@@ -0,0 +1,360 @@
...
+    std::copy(peers.begin(), peers.end(), std::back_inserter(peer_index));
+    std::sort(
+            peer_index.begin(), peer_index.end(), [](const auto& peer1, const auto& peer2)
+            { return peer1.get().uuid < peer2.get().uuid; }
+    );
+
+}
+
+void
+pbft::start()
```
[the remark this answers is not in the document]
isabelsavannah: Isn't this only called from main?

pbft/pbft_operation.hpp
```diff
@@ -0,0 +1,80 @@
...
+    };
+
+
+    class pbft_operation
+    {
+    public:
+
+        pbft_operation(uint64_t view, uint64_t sequence, pbft_request msg, const peers_list_t& peers);
+
+        static request_hash_t request_hash(const pbft_request& req);
```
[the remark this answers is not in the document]
isabelsavannah: When pbft_operation starts representing something on disk would be the time to introduce that, I think.

pbft/pbft.hpp
```diff
@@ -0,0 +1,95 @@
...
+        void unwrap_message(const bzn::message& json, std::shared_ptr<bzn::session_base> session);
+        bzn::message wrap_message(const pbft_msg& message);
+        pbft_msg common_message_setup(const pbft_operation& op);
+        void broadcast(const pbft_msg& message);
+
+        // Using 1 as first value here to distinguish from default value of 0 in protobuf
+        uint64_t view = 1;
+        uint64_t next_issued_sequence_number = 1;
+
+        uint64_t low_water_mark = 1;
```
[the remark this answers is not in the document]
isabelsavannah: none of these are constant in a full implementation

pbft/pbft_operation.cpp
```diff
@@ -0,0 +1,111 @@
...
+void
+pbft_operation::record_commit(const pbft_msg& commit)
+{
+    this->commits_seen.insert(commit.sender());
+}
+
+bool
+pbft_operation::is_committed()
+{
+    size_t f = (this->peers.size() - 1) / 3;
```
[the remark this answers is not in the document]
isabelsavannah: size_t is required to be at least 16 bits, and there are a million other things that will break before we have that many peers.

I have pulled out the definition of f, but the constants don't really have names, so unless you want "const int THREE = 3"...

isabelsavannah:
I am well aware that "if(condition){return A;}else{return B;}" is equivalent to "if(condition){return A;} return B;" and when I have chosen the former it usually for the sake of readability. Is this really that contentious of a style issue?

ebruck:
I've found temp variables for readability are sometimes needed, but not in the cases I commented on.
Also, reducing else cases does make things more readable IMHO. I've also seen them flagged by static analysis tools, which we will be adding in the near future.

pbft/pbft_base.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+namespace bzn
+{
+    using client_t = std::string; //placeholder
+
+    class pbft_base
+    {
+    public:
+        virtual void start() = 0;
+
```
ebruck: you are missing a virtual destructor.

Please add:

virtual ~pbft_base() = default;

pbft/pbft.cpp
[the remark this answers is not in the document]
isabelsavannah: Then wouldn't it be better to throw an error, rather than silently ignoring subsequent calls?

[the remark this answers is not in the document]
isabelsavannah: Yeah, that's my thinking. If we accidentally try to start one of these modules twice, then that's likely something very and obviously wrong - and it'd be easier to debug if it was an obvious crash rather than silently ignored at runtime.

pbft/pbft.cpp
```diff
@@ -38,46 +38,49 @@ pbft::pbft(
...
 void
 pbft::unwrap_message(const bzn::message& json, std::shared_ptr<bzn::session_base> /*session*/)
 {
     pbft_msg msg;
-    msg.ParseFromString(boost::beast::detail::base64_decode(json["pbft-data"].asString()));
-
-    this->handle_message(msg);
+    if (msg.ParseFromString(boost::beast::detail::base64_decode(json["pbft-data"].asString())))
+    {
+        this->handle_message(msg);
```
ebruck: I would return here and lose the else.

pbft/pbft.cpp
```diff
@@ -150,9 +153,8 @@ pbft::handle_request(const pbft_msg& msg)
 
     //TODO: keep track of what requests we've seen based on timestamp and only send preprepares once - KEP-329
 
-    uint64_t request_view = this->view;
     uint64_t request_seq = this->next_issued_sequence_number++;
```
ebruck: const uint64_t

pbft/pbft.cpp
```diff
@@ -0,0 +1,363 @@
...
+    {
+        LOG(error) << "Got invalid message " << json.toStyledString().substr(0, MAX_MESSAGE_SIZE);
+    }
+}
+
+void
+pbft::handle_message(const pbft_msg& msg)
+{
+
+    LOG(debug) << "Recieved message: " << msg.ShortDebugString().substr(0, MAX_MESSAGE_SIZE);
```
ebruck: spelling error

pbft/pbft.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+// Find this node's record of an operation (creating a new record for it if this is the first time we've heard of it)
+pbft_operation&
+pbft::find_operation(const pbft_msg& msg)
+{
+    return this->find_operation(msg.view(), msg.sequence(), msg.request());
+}
+
+pbft_operation&
+pbft::find_operation(const uint64_t& view, const uint64_t& sequence, const pbft_request& request)
```
ebruck: I don't believe you gain anything by passing POD as a const reference over a copy.
More style I guess, but I could just do pass by value and mark the variable const in the implementation.

pbft/pbft.cpp
```diff
@@ -0,0 +1,363 @@
...
+        LOG(error) << "Ignoring client request because I am not the leader";
+        // TODO - KEP-327
+        return;
+    }
+
+    //TODO: conditionally discard based on timestamp - KEP-328
+
+    //TODO: keep track of what requests we've seen based on timestamp and only send preprepares once - KEP-329
+
+    uint64_t request_seq = this->next_issued_sequence_number++;
```
ebruck: const uint64_t?

pbft/pbft_service.cpp
```diff
@@ -0,0 +1,47 @@
...
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+#include "pbft/pbft_service.hpp"
+
+using namespace bzn;
+
+void
+pbft_service::commit_request(uint64_t sequence, const pbft_request& request)
+{
+    waiting_requests[sequence] = request;
```
ebruck: this->

pbft/pbft_service.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+    while (waiting_requests.count(this->next_request_sequence) > 0)
+    {
+        const pbft_request& req = waiting_requests[sequence];
```
ebruck: this->

pbft/pbft_service.hpp
```diff
@@ -0,0 +1,41 @@
...
+    {
+    public:
+        void commit_request(uint64_t sequence, const pbft_request& request) override;
+
+        const pbft_request& get_last_request();
+        uint64_t applied_requests_count();
+
+    private:
+        uint64_t next_request_sequence = 1;
+        std::map<uint64_t, pbft_request> waiting_requests;
```
ebruck: any reason not to use unordered_map?

proto/pbft.proto
```diff
@@ -0,0 +1,30 @@
...
+    uint64 sequence = 3;
+    // TODO: Most messages should contain only the hash of the request - KEP-344
+    pbft_request request = 4;
+
+    // For prepare, commit, etc messages where the recipient needs to count the senders. This is the sender's uuid.
+    string sender = 5;
+}
+
+enum pbft_msg_type {
+    PBFT_MSG_TYPE_UNDEFINED = 0;
```
ebruck: _TYPE_ isn't really needed in the name

pbft/pbft.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+    return msg;
+}
+
+void
+pbft::do_preprepare(pbft_operation& op)
+{
+    LOG(debug) << "Doing preprepare for operation " << op.debug_string();
+
+    pbft_msg msg = this->common_message_setup(op);
```
ebruck: perhaps send in message type to helper function so you can remove all of the msg.set_type calls?

pbft/pbft_service.hpp
isabelsavannah: From what I'm seeing, no strong reason either way. This implementation of pbft_service is mostly only for debugging pbft, anyway.

PR #85, opened 2018-07-20
KEP-489: Removing the .state file functionality as it is no longer re…

rnistuk:
…quired and was causing nodes to fail on restart

### DIFF
--- a/raft/raft.hpp
+++ b/raft/raft.hpp
@@ -120,12 +120,12 @@ namespace bzn
         void get_raft_timeout_scale();
 
         std::string entries_log_path();
+
         std::string state_path();
-        void save_state();
-        void load_state();
+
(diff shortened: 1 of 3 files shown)


raft/raft.hpp
```diff
@@ -120,12 +120,12 @@ namespace bzn
...
         std::string entries_log_path();
+
         std::string state_path();
-        void save_state();
-        void load_state();
+
         void import_state_files();
-        void create_state_files(const std::string& log_path, const bzn::peers_list_t& peers);
-        bool state_files_exist();
+        void create_dat_file(const std::string& log_path, const bzn::peers_list_t& peers);
```
ebruck: What's a dat?  Digital Audio Tape? :)

PR #90, opened 2018-07-27
KEP-433: Load storage from log on restart

rnistuk:
Fixed the bug where the log was not being read into storage when the node was restarted.

### DIFF
--- a/raft/raft.cpp
+++ b/raft/raft.cpp
@@ -58,7 +58,7 @@ raft::raft(
     this->setup_peer_tracking(peers);
     this->get_raft_timeout_scale();
 
-    if(!boost::filesystem::exists(this->entries_log_path()))
+    if (!boost::filesystem::exists(this->entries_log_path()))
     {
         this->create_dat_file(this->entries_log_path(), peers);
     }
@@ -66,7 +66,7 @@ raft::raft(
     this->raft_log = std::make_shared<bzn::raft_log>(this->entries_log_path(), maximum_raft_storage);
 
     // KEP-112 Bail if the raft storage exceeds the max storage
-    if(this->raft_log->maximum_storage_exceeded())
+    if (this->raft_log->maximum_storage_exceeded())
     {
         LOG(error) << MSG_ERROR_MAXIMUM_STORAGE_EXCEEDED;
         throw std::runtime_error(MSG_ERROR_MAXIMUM_STORAGE_EXCEEDED);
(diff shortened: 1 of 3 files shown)


raft/raft.cpp
```diff
@@ -867,27 +867,44 @@ raft::get_leader()
...
     {
         if(log_entry.entry_type == bzn::log_entry_type::database)
         {
-            const auto command = log_entry.msg["cmd"].asString();
-            const auto db_uuid = log_entry.msg["db-uuid"].asString();
-            const auto key = log_entry.msg["data"]["key"].asString();
-            if (command == "create")
+            bzn_msg msg;
+
+            if(!msg.ParseFromString(boost::beast::detail::base64_decode(log_entry.msg["msg"].asString())))
```
ebruck: space after if

raft/raft.cpp
```diff
@@ -867,27 +867,44 @@ raft::get_leader()
...
             {
-                storage->create(db_uuid, key, log_entry.msg["data"]["value"].asString());
+                LOG(error) << "Failed to decode message: " << boost::beast::detail::base64_decode(log_entry.msg["msg"].asString()).substr(0,MAX_MESSAGE_SIZE) << "...";
+                continue;
             }
-            else if (command == "update")
+
+            bzn::uuid_t uuid = msg.db().header().db_uuid();
```
ebruck: const bzn::uuid_t

raft/test/raft_test.cpp
```diff
@@ -189,6 +191,87 @@ namespace
         peer["uuid"] = TEST_NODE_UUID;
         return peer;
     }
+
+
+    database_header*
```
ebruck: I'm surprised protobuf doesn't use smart pointers.

raft/raft.cpp
rnistuk: Right, good catch.

raft/test/raft_test.cpp
rnistuk: Me too, that threw me for a loop. I didn't think to look at the doc for the "set_allocated_" type functions, so when I used &variable on variables allocated on the stack, when a bzn_msg went out of scope, there was an exception. To make that work you need to match each "set_allocated_" with a "release_", otherwise, do pointers and allow bzn_msg to clean up for you by calling the release...

raft/raft.cpp
rnistuk: SON OF A... ok ok, I missed this one .. my bad... As punishment for my transgression I made the change to all the other offending if statements.

PR #91, opened 2018-07-30
KEP-299 Implement Node State Request API Function

ebruck:


status/test/status_test.cpp
```diff
@@ -0,0 +1,96 @@
...
+    bzn::message request;
+    request["bzn-api"] = "status";
+    request["cmd"] = "state";
+    request["transaction-id"] = 85746;
+
+    EXPECT_CALL(*mock_status_provider, get_status()).Times(2).WillRepeatedly(Invoke(
+        []()
+        {
+            bzn::message status;
+            status["line"] = __LINE__;
```
rnistuk: Is there a reason to set these values? ("line" and "file")

ebruck: just something to output

PR #101, opened 2018-08-11
KEP-536 Moved the check for max storage exceeded up the stack from ap…

rnistuk:
…pend_log_unsafe to handle_ws_raft_messages

The problem was that the append_log_unsafe method was only being called by the leader. So if a follower had exceeded it's max storage it would happily append messages using a different path (which I think may be a problem we should look at, but since we are doing away with raft...)

I combined the duplicated code where we bail on exceeded storage into a nice method and called it on raft instantiation and at the start of handle_ws_raft_messages before we attempt to do any work.

NOTE: I accidentally put KEP-563 in the git log, it should be KEP-536.

### DIFF
--- a/raft/raft.cpp
+++ b/raft/raft.cpp
@@ -65,12 +65,7 @@ raft::raft(
 
     this->raft_log = std::make_shared<bzn::raft_log>(this->entries_log_path(), maximum_raft_storage);
 
-    // KEP-112 Bail if the raft storage exceeds the max storage
-    if (this->raft_log->maximum_storage_exceeded())
-    {
-        LOG(error) << MSG_ERROR_MAXIMUM_STORAGE_EXCEEDED;
-        throw std::runtime_error(MSG_ERROR_MAXIMUM_STORAGE_EXCEEDED);
-    }
+    this->shutdown_on_exceeded_max_storage(true);
 }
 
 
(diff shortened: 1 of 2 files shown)


raft/raft.cpp
```diff
@@ -65,12 +65,7 @@ raft::raft(
 
     this->raft_log = std::make_shared<bzn::raft_log>(this->entries_log_path(), maximum_raft_storage);
 
-    // KEP-112 Bail if the raft storage exceeds the max storage
-    if (this->raft_log->maximum_storage_exceeded())
-    {
-        LOG(error) << MSG_ERROR_MAXIMUM_STORAGE_EXCEEDED;
-        throw std::runtime_error(MSG_ERROR_MAXIMUM_STORAGE_EXCEEDED);
-    }
+    this->bail_on_exceeded_max_storage(true);
```
ebruck: Can we please call this: 

shutdown_on_exceeded_max_storage?

raft/raft.cpp
```diff
@@ -1184,3 +1174,14 @@ raft::get_status()
...
     return status;
 }
+
+
+void
+raft::bail_on_exceeded_max_storage(bool do_throw)
+{
+    if (this->raft_log->maximum_storage_exceeded())
+    {
+        LOG(error) << MSG_ERROR_MAXIMUM_STORAGE_EXCEEDED;
```
ebruck: this->do_throw

raft/raft.cpp
rnistuk: Thanks! Made the change.

raft/raft.cpp
rnistuk: I wanted to leave one as a throw simply because I was able to unit test the functionality easily.

rnistuk: Can we catch a SIGINT? Maybe with signal?

PR #103, opened 2018-08-14
KEP 333, 521: Failure detection for pbft

isabelsavannah:


mocks/mock_pbft_failure_detector.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+#pragma once
+
+#include <pbft/pbft_failure_detector_base.hpp>
+#include <gmock/gmock.h>
+#include <functional>
+
+namespace bzn
+{
+    class Mockpbft_failure_detector_base : public pbft_failure_detector_base
```
ebruck: Was this handcrafted? If so then you may want to in the future use gmock_gen.y located:

./googletest/src/googletest/googlemock/scripts/generator/gmock_gen.py

pbft/pbft.cpp
```diff
@@ -80,6 +82,18 @@ pbft::start()
                 this->audit_heartbeat_timer->expires_from_now(heartbeat_interval);
                 this->audit_heartbeat_timer->async_wait(
                         std::bind(&pbft::handle_audit_heartbeat_timeout, shared_from_this(), std::placeholders::_1));
+
+                auto weak_this = this->weak_from_this();
+                this->failure_detector->register_failure_handler(
```
ebruck: With c++17 you can avoid the copy...

[auto weak_this=this->weak_from_this()]()
{
...

pbft/pbft.cpp
```diff
@@ -195,9 +215,9 @@ pbft::handle_request(const pbft_msg& msg)
     //TODO: keep track of what requests we've seen based on timestamp and only send preprepares once - KEP-329
 
     const uint64_t request_seq = this->next_issued_sequence_number++;
-    pbft_operation& op = this->find_operation(this->view, request_seq, msg.request());
+    std::shared_ptr<pbft_operation> op = this->find_operation(this->view, request_seq, msg.request());
```
ebruck: I'd move this->find_operation() into do_preprepare call or use auto.

pbft/pbft.cpp
```diff
@@ -218,11 +238,11 @@ pbft::handle_preprepare(const pbft_msg& msg)
     }
     else
     {
-        pbft_operation& op = this->find_operation(msg);
-        op.record_preprepare();
+        std::shared_ptr<pbft_operation> op = this->find_operation(msg);
```
ebruck: auto?

pbft/pbft.cpp
```diff
@@ -234,55 +254,57 @@ pbft::handle_prepare(const pbft_msg& msg)
...
+    this->maybe_advance_operation_state(std::move(op));
 }
 
 void
 pbft::handle_commit(const pbft_msg& msg)
 {
 
     // Commit messages are never rejected, assuming  the sanity checks passed
-    pbft_operation& op = this->find_operation(msg);
+    std::shared_ptr<pbft_operation> op = this->find_operation(msg);
```
ebruck: auto?

pbft/pbft_service.cpp
```diff
@@ -16,18 +16,28 @@
...
+pbft_service::commit_operation(std::shared_ptr<pbft_operation> operation)
 {
-    this->waiting_requests[sequence] = request;
+    uint64_t sequence = operation->sequence;
+    this->waiting_requests[sequence] = std::move(operation);
 
     while (this->waiting_requests.count(this->next_request_sequence) > 0)
     {
-        const pbft_request& req = waiting_requests[sequence];
+        std::shared_ptr<pbft_operation>& op = waiting_requests[this->next_request_sequence];
```
ebruck: This could be dangerous if the original shared_ptr becomes invalid?
I would remove the reference.

pbft/pbft_service.hpp
```diff
@@ -26,15 +27,18 @@ namespace bzn
...
+
+        void commit_operation(std::shared_ptr<pbft_operation> operation) override;
 
         const pbft_request& get_last_request();
         uint64_t applied_requests_count();
 
     private:
         uint64_t next_request_sequence = 1;
-        std::map<uint64_t, pbft_request> waiting_requests;
+        std::map<uint64_t, std::shared_ptr<pbft_operation>> waiting_requests;
```
ebruck: any reason not to use an unordered map?

pbft/pbft.cpp
```diff
@@ -293,49 +315,49 @@ pbft::common_message_setup(const pbft_operation& op, pbft_msg_type type)
 }
 
 void
-pbft::do_preprepare(pbft_operation& op)
+pbft::do_preprepare(const std::shared_ptr<pbft_operation>& op)
```
ebruck: I'd avoid references to shared pointers unless you are absolutely sure the caller of this will never lose its ref count and it goes to zero.

pbft/pbft.cpp
isabelsavannah: Sorry, how? I'm not following.

pbft/pbft.cpp
isabelsavannah: I have been consistent about that. We make a real copy of the pointer in handle_[message type], then everything from there uses references to that copy. I wanted to avoid copying it for all the small functions.

pbft/pbft.cpp
ebruck: oops... it stripped my formatting for some reason:

```
[auto weak_this = this->weak_from_this()]()
{

}
```

pbft/pbft_service.hpp
isabelsavannah: Not really. Last time I googled it, I think I found that the performance difference wasn't as compelling or consistent as you'd expect.

pbft/pbft.cpp
isabelsavannah: You sure? I get 
```/home/isabel/swarmDB/pbft/pbft.cpp: In lambda function:
/home/isabel/swarmDB/pbft/pbft.cpp:87:26: error: expected identifier before ‘auto’
                         [auto weak_this = this->weak_from_this()]()
```
GCC 8.0.1

ebruck: sorry auto isn't allowed. My mistake. But I would put it there anyway.

pbft/pbft.cpp
ebruck: If you're not sharing ownership and the original holder of the reference doesn't go out of scope then this is fine.

pbft/pbft.cpp
isabelsavannah: It doesn't like that either
```
/home/isabel/swarmDB/pbft/pbft.cpp: In lambda function:
/home/isabel/swarmDB/pbft/pbft.cpp:88:26: error: capture of non-variable ‘std’ 
                         [std::weak_ptr<pbft> weak_this = this->weak_from_this()]()
                          ^~~
<built-in>: note: ‘namespace std { }’ declared here
/home/isabel/swarmDB/pbft/pbft.cpp:88:29: error: expected ‘,’ before ‘::’ token
                         [std::weak_ptr<pbft> weak_this = this->weak_from_this()]()
                             ^~
```

ebruck: gah... my brain is turning to mush... drop the type and use like:

```
[weak_this = this->weak_from_this()]()
{
}
```

ebruck:
Please wait until coverage is enabled. Your PR has moved us below 87%

PR #106, opened 2018-08-15
KEP-190 Whitelist implementation

rnistuk:
Implements simple whitelist checking against an ethereum contract that Mehdi has created. The functionality is currently off by default, but can be turned on by setting "enable_whitelist" to true in the config file.

I'll be talking to Mehdi to add 100 test uuids to the whitelist. Currently, the whitelist contains just one: "12345678-0900-0000-0000-000000000000"


rnistuk:
I also DRY'd things up a bit in raft.cpp by consolidating some error reporting code into a small method.

raft/raft.cpp
```diff
@@ -39,16 +40,18 @@ using namespace bzn;
...
-    , uuid(std::move(uuid))
-    , node(std::move(node))
-    , state_dir(std::move(state_dir))
+        std::shared_ptr<bzn::asio::io_context_base> io_context,
+        std::shared_ptr<bzn::node_base> node,
+        const bzn::peers_list_t& peers,
+        bzn::uuid_t uuid,
+        const std::string state_dir, size_t maximum_raft_storage,
+        bool whitelist_enabled
+        ) :
```
ebruck: this should be : on a new line and comma's underneath it 

like:

)
: timer()
, uuid()
, node()

etc.

rnistuk: Cool, I'll make the change.

PR #110, opened 2018-08-21
KEP-548: Add OpenSSL library to project

rnistuk:
Just added the find package line that includes the OpenSSL package.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,8 @@ if(Boost_FOUND)
 endif()
 
 FIND_PACKAGE(CURL REQUIRED)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -56,6 +56,8 @@ if(Boost_FOUND)
 endif()
 
 FIND_PACKAGE(CURL REQUIRED)
+FIND_PACKAGE(OpenSSL REQUIRED)
```
ebruck: We have standardized on lowercase cmake. Please fix these two lines.

rnistuk: ok

PR #116, opened 2018-09-04
KEP-508 secure the whitelist functionality

rnistuk:
Black list and verifying uuid's against signatures created with Bluzelle owned private keys.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,7 +150,7 @@ The configuration file is a JSON format file, as seen in the following example:
         "logfile_rotation_size" : "64K",
         "logfile_max_size" : "640K",
         "debug_logging" : false,
-        "security_enabled" : false 
+        "peer_validation_enabled" : false 
     }
 
 where the properties are:
@@ -168,7 +168,7 @@ where the properties are:
 - "logfile_rotation_size" - approximate size of log file must be before rotation (default: 64K)
 - "max_storage" - the approximate maximum limit for the storage that SwarmDB will use in the current instance (default: 2G)
 - "uuid" - the universally unique identifier that this instance of SwarmDB will use to uniquely identify itself.
-- "security_enabled" - set this to true to enable blacklisting and uuid signature verification
+- "peer_validation_enabled" - set this to true to enable blacklisting and uuid signature verification
 
 All size entries use the same notation as storage: B, K, M, G & T or none
 (bytes)
@@ -651,7 +651,7 @@ swarm until you add it to the swarm.
 
 Create your add_peer JSON object, and use wscat to send it to the swarm leader,
 note that the signature object is only required for swarms whose nodes have set
-the security_enabled object to true in thier config files:
+the peer_validation_enabled object to true in thier config files:
 
     $ wscat -c  http://<leader-address>:<port>
     connected (press CTRL+C to quit)
--- a/README.md
+++ b/README.md
@@ -149,7 +149,8 @@ The configuration file is a JSON format file, as seen in the following example:
         "logfile_dir" : "logs/",
         "logfile_rotation_size" : "64K",
         "logfile_max_size" : "640K",
-        "debug_logging" : false
+        "debug_logging" : false,
+        "security_enabled" : false 
     }
 
 where the properties are:
@@ -167,6 +168,7 @@ where the properties are:
 - "logfile_rotation_size" - approximate size of log file must be before rotation (default: 64K)
 - "max_storage" - the approximate maximum limit for the storage that SwarmDB will use in the current instance (default: 2G)
 - "uuid" - the universally unique identifier that this instance of SwarmDB will use to uniquely identify itself.
+- "security_enabled" - set this to true to enable blacklisting and uuid signature verification
 
 All size entries use the same notation as storage: B, K, M, G & T or none
 (bytes)
@@ -612,7 +614,8 @@ following JSON objects:
                 "http_port":<HTTPPORT>,
                 "name":"<NODE-NAME>",
                 "port":<PORT>,
-                "uuid":"<UUID>"
+                "uuid":"<UUID>",
+                "signature" : ["<signature line>", ..., "<last signature line>"]
             }
         }
     }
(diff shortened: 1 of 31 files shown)


README.md
```diff
@@ -149,7 +149,8 @@ The configuration file is a JSON format file, as seen in the following example:
         "logfile_dir" : "logs/",
         "logfile_rotation_size" : "64K",
         "logfile_max_size" : "640K",
-        "debug_logging" : false
+        "debug_logging" : false,
+        "security_enabled" : false
```
ebruck: wouldn't a name like "peer_validation" be better?

bootstrap/bootstrap_peers.cpp
```diff
@@ -103,6 +105,18 @@ bootstrap_peers::initialize_peer_list(const Json::Value& root, bzn::peers_list_t
             continue;
         }
 
+        if (this->is_security_enabled())
+        {
+            // At this point we cannot validate uuids as we do not have
+            // signatures for all of them, so we simply ignore blacklisted
+            // uuids
+            if(bzn::utils::blacklist::is_blacklisted(uuid))
```
ebruck: space after if

raft/raft.cpp
```diff
@@ -423,6 +424,33 @@ raft::send_session_error_message(std::shared_ptr<bzn::session_base> session, con
...
+    // uuid's are signed as text files, so we must append the end of line character
+    const auto uuid{peer["uuid"].asString() + "\x0a"};
+    if (!peer.isMember("signature")  || peer["signature"].asString().empty())
+    {
+        this->send_session_error_message(session, ERROR_INVALID_SIGNATURE);
+        return false;
+    }
+
+    const auto signature{peer["signature"].asString()};
+    if(!bzn::utils::crypto::verify_signature(bzn::utils::crypto::retrieve_bluzelle_public_key_from_contract(), signature, uuid))
```
ebruck: space after if

utils/crypto.cpp
```diff
@@ -0,0 +1,236 @@
...
+     * This function is used by base_64_decode to determine the amount of memory required
+     * for the binary buffer represented by the base64 encoded string.
+     *
+     * @param base_64_input base64 encoded string
+     * @return a size_t containing the number of bytes expected in the decoded string.
+    */
+    size_t
+    calculate_decode_length(const std::string& base_64_input)
+    {
+        const auto len{base_64_input.size()};
```
ebruck: change auto to size_t

utils/crypto.cpp
```diff
@@ -0,0 +1,236 @@
...
+    /**
+     * This function is used by base_64_decode to determine the amount of memory required
+     * for the binary buffer represented by the base64 encoded string.
+     *
+     * @param base_64_input base64 encoded string
+     * @return a size_t containing the number of bytes expected in the decoded string.
+    */
+    size_t
+    calculate_decode_length(const std::string& base_64_input)
+    {
```
ebruck: What if input is empty?

utils/crypto.cpp
```diff
@@ -0,0 +1,236 @@
...
+        auto padding = (base_64_input[len - 1] == '=' && base_64_input[len - 2] == '=') ? 2 : 1;
+        return (len * 3) / 4 - padding;
+    }
+
+
+    void
+    compose_and_log_OpenSSL_error(const std::string& failing_function_name)
+    {
+        char buffer[126]{0};
+        ERR_error_string_n(ERR_get_error(), buffer, 126);
```
ebruck: I think you can do sizeof(buffer)

utils/crypto.cpp
```diff
@@ -0,0 +1,236 @@
...
+     * Given the public key string in PEM format, this function will return the
+     * RSA public key structure that OpenSSL functions use.
+     *
+     * @param key a const std::string reference containing the public key in PEM format
+     * @return a pointer to the RSA structure contqining the public key
+     */
+    RSA* create_public_RSA(const std::string& key)
+    {
+        std::unique_ptr<BIO,std::function<void(BIO*)>> keybio(BIO_new_mem_buf((void*)(key.c_str()), -1), BIO_free);
+        if (nullptr==keybio)
```
ebruck: if (keybio) is all you need.

utils/crypto.cpp
```diff
@@ -0,0 +1,236 @@
...
+    {
+        // TODO retrieve the public key from the contract
+        return temporary_public_pem;
+    }
+
+
+    int
+    base64_decode(const std::string& base64_message, std::vector<unsigned char>& decoded_message)
+    {
+        const auto decoded_length = calculate_decode_length(base64_message);
```
ebruck: this isn't the kind of places auto is used for

utils/crypto.cpp
```diff
@@ -0,0 +1,236 @@
...
+        // it probably doesn't.
+        std::string nonconst{base64_message};
+        BIO* bio = BIO_new_mem_buf(nonconst.data(), -1);
+
+        BIO* b64 = BIO_new(BIO_f_base64());
+        bio = BIO_push(b64, bio);
+
+        BIO_set_flags(bio, BIO_FLAGS_BASE64_NO_NL); //Do not use newlines to flush buffer
+
+        const auto length = BIO_read(bio, decoded_message.data(), static_cast<int>(base64_message.size()));
```
ebruck: same here

README.md
rnistuk: Yes. Thanks for the suggestion!

rnistuk: OK, did that refactoring and pushed.

bootstrap/bootstrap_peers.cpp
rnistuk: Right! OK, as penance I have scoured the code base and corrected this everywhere I found it.

raft/raft.cpp
rnistuk: Fixed. ibid.

utils/crypto.cpp
rnistuk: Added a test to the first line of the function that returns 0 if base_64_input is empty.

Also went back up the stack to the base64_decode function to return a failure and throw up a log message if the input that is supposed to contain a base64 encoded string is empty.

Added a unit test as well.

utils/crypto.cpp
rnistuk: Yes, I can. Done.

utils/crypto.cpp
rnistuk: Right, I'll stop using auto for pod.

utils/crypto.cpp
rnistuk: Changed auto to int

PR #118, opened 2018-09-06
fixing the openssl issue

rnistuk:


ebruck:
Please use the **KEP-XYZ description** commit message format.

PR #126, opened 2018-09-15
KEP-599: Daemon cryptographic identity

isabelsavannah:
Dameon uses own public key as uuid, script to generate such key

### DIFF
--- a/options/options.cpp
+++ b/options/options.cpp
@@ -16,6 +16,7 @@
 #include <boost/lexical_cast.hpp>
 #include <regex>
 #include <cstdint>
+#include <utils/crypto.hpp>
 
 using namespace bzn;
 using namespace bzn::option_names;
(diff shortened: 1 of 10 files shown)


utils/crypto.cpp
```diff
@@ -34,6 +34,7 @@
 #include <stdexcept>
 #include <iostream>
 #include <memory>
+#include <stdio.h>
```
ebruck: please use cstdio

utils/crypto.cpp
```diff
@@ -256,4 +257,34 @@ namespace bzn::utils::crypto
 
         return ret_val & authentic;
     }
+
+    std::string read_pem_file(std::string filename, std::string expected_type)
```
ebruck: const std::string& filename, const std::string& excepted_type

options/options.cpp
```diff
@@ -118,8 +119,13 @@ options::get_bootstrap_peers_url() const
...
 options::get_uuid() const
 {
-    //TODO: Remove this
-    return this->raw_opts.get<std::string>(NODE_UUID);
+    if (this->raw_opts.has(NODE_UUID))
+    {
+        return this->raw_opts.get<std::string>(NODE_UUID);
+    }
+
+    std::string pubkey_raw = bzn::utils::crypto::read_pem_file(this->raw_opts.get<std::string>(NODE_PUBKEY_FILE), "PUBLIC KEY");
```
ebruck: const std::string

options/test/options_test.cpp
```diff
@@ -42,8 +43,14 @@ namespace
         "  \"logfile_max_size\" : \"1M\","
         "  \"logfile_rotation_size\" : \"2M\","
         "  \"logfile_dir\" : \".\","
-        "  \"http_port\" : 80"
-        "}";
+        "  \"http_port\" : 80";
+
+    const std::string DEFAULT_CONFIG_DATA = "{" + DEFAULT_CONFIG_CONTENT + "}";
+
+    std::string compose_config_data(const std::string &a, const std::string &b)
```
ebruck: return on separate line
const std::string& a, const std::string& b   <--- & with type please

PR #131, opened 2018-09-25
KEP-590: Chaos Module

isabelsavannah:


node/test/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
 set(test_srcs node_test.cpp session_test.cpp)
-set(test_libs node proto protobuf.a)
+set(test_libs node proto options protobuf.a)
```
ebruck: remove options

chaos/chaos.cpp
```diff
@@ -0,0 +1,125 @@
...
+    if (result)
+    {
+        LOG(debug) << "Chaos module dropping message";
+    }
+
+    return result;
+}
+
+void
+chaos::reschedule_message(std::function<void()> callback) const
```
ebruck: another callback type

node/session.cpp
```diff
@@ -24,11 +24,12 @@ namespace
 using namespace bzn;
 
 
-session::session(std::shared_ptr<bzn::asio::io_context_base> io_context, const bzn::session_id session_id, std::shared_ptr<bzn::beast::websocket_stream_base> websocket, const std::chrono::milliseconds& ws_idle_timeout)
+session::session(std::shared_ptr<bzn::asio::io_context_base> io_context, const bzn::session_id session_id, std::shared_ptr<bzn::beast::websocket_stream_base> websocket, std::shared_ptr<bzn::chaos_base> chaos, const std::chrono::milliseconds& ws_idle_timeout)
```
ebruck: There's no way to disable chaos?

isabelsavannah: chaos itself does the check; is_message_dropped and is_message_delayed always return false if its disabled

PR #132, opened 2018-09-25
KEP-596 Replace existing storage in memory database with RocksDB

ebruck:


storage/mem_storage.hpp
```diff
@@ -16,42 +16,32 @@
...
-        storage_base::result load(const std::string& path) override;
-
         std::vector<std::string> get_keys(const bzn::uuid_t& uuid) override;
 
         bool has(const bzn::uuid_t& uuid, const  std::string& key) override;
 
         std::size_t get_size(const bzn::uuid_t& uuid) override;
 
     private:
-        friend class boost::serialization::access;
```
rnistuk: Nice that we can get rid of the serialization stuff.

swarm/main.cpp
```diff
@@ -268,6 +284,7 @@ main(int argc, const char* argv[])
     }
     catch(std::exception& ex)
     {
+        LOG(fatal) << ex.what();
```
rnistuk: Cool!

PR #136, opened 2018-10-03
EP-560 Daemon can get bluzelle public key from Solidity contract on E…

rnistuk:
…thereum

Also snuck in the fix for the protobuf issue on macOS.

### DIFF
--- a/utils/crypto.cpp
+++ b/utils/crypto.cpp
@@ -35,9 +35,16 @@
 #include <iostream>
 #include <memory>
 #include <cstdio>
+#include <curl/curl.h>
+#include <boost/format.hpp>
+#include <iomanip>
 
 namespace
 {
+    const std::string ROPSTEN_URL{"https://ropsten.infura.io"};
+    const std::string REQUEST_BASE              {R"({"jsonrpc":"2.0","method":"eth_call","params":[{"to":"0x492CCD1eAeCDA0262e3AAeF8445F3F731a30AfbB","data": "%s" },"latest"],"id":1})"};
+    const std::string GET_KEY_SIZE              {"0xcf5d53f1"};
+    const std::string GET_KEY_CHUNK             {"0x96ce93e2"};
     /**
      * NOTE This is temporary
      * We shall use this only until we can get the public key from an Etherium
@@ -61,8 +68,6 @@ namespace
             "MwoaYGgrejBmlvZ5UiFw+xMCAwEAAQ==\n"
             "-----END PUBLIC KEY-----"
     };
-
-
 
 
     /**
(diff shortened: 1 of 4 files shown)


utils/crypto.cpp
```diff
@@ -178,6 +183,111 @@ namespace
             return false;
         }
     }
+
+
+    size_t
+    write_function(void* contents,size_t size, size_t nmemb, void* userp)
```
ebruck: nmemb?

utils/crypto.cpp
```diff
@@ -178,6 +183,111 @@ namespace
...
+    {
+        reinterpret_cast<std::string*>(userp)->append(reinterpret_cast<char*>(contents), size * nmemb);
+        return size * nmemb;
+    }
+
+
+    bzn::json_message
+    get_curl_response(const std::string& function_call, const std::string& url = ROPSTEN_URL)
+    {
+        const std::string MSG_ERROR_CURL{"curl_easy_perform() failed: "};
```
ebruck: I would put this outside so it's initialized only once.

utils/crypto.cpp
```diff
@@ -178,6 +183,111 @@ namespace
...
+        }
+        return ss.str();
+    }
+
+
+    // TODO rename hex_string_to_string
+    std::string
+    hex_string_to_string(const std::string& hex_string)
+    {
+        std::string public_key{""};
```
ebruck: init not required

utils/crypto.cpp
```diff
@@ -186,8 +296,20 @@ namespace bzn::utils::crypto
     std::string
     retrieve_bluzelle_public_key_from_contract()
     {
+        std::string public_key_hex{""};
         // TODO retrieve the public key from the contract
-        return temporary_public_pem;
+        const size_t key_size = get_public_pem_size();
```
ebruck: why the temp? You don't it elsewhere.

utils/crypto.cpp
```diff
@@ -186,8 +296,20 @@ namespace bzn::utils::crypto
     std::string
     retrieve_bluzelle_public_key_from_contract()
     {
+        std::string public_key_hex{""};
```
ebruck: unnecessary init

utils/crypto.cpp
```diff
@@ -178,6 +183,111 @@ namespace
...
+        curl_easy_setopt(curl.get(), CURLOPT_POSTFIELDS, post_fields.c_str());
+
+        curl_easy_setopt(curl.get(), CURLOPT_POSTFIELDSIZE, -1L);
+        curl_easy_setopt(curl.get(), CURLOPT_URL, url.c_str());
+        curl_easy_setopt(curl.get(), CURLOPT_WRITEFUNCTION, reinterpret_cast<void*>(write_function));
+
+        CURLcode res = curl_easy_perform(curl.get());
+        if (res != CURLE_OK)
+        {
+            throw (std::runtime_error(MSG_ERROR_CURL + curl_easy_strerror(res)));
```
ebruck: Is this truly a fatal error?

utils/crypto.cpp
```diff
@@ -178,6 +183,111 @@ namespace
...
+        }
+
+        bzn::json_message response;
+        Json::Reader reader;
+
+        if (!reader.parse(readBuffer, response))
+        {
+            LOG(error) << "Unable to parse response from Ropsten - could not get key size";
```
ebruck: If the above is a fatal error then why don't we throw here?

utils/crypto.cpp
rnistuk: OK.  Done.

utils/crypto.cpp
rnistuk: Removed.

utils/crypto.cpp
rnistuk: Thanks, good catch. Removed.

utils/crypto.cpp
rnistuk: Removed.

utils/crypto.cpp
rnistuk: Probably not. I think I can just do what I do in the net block. I'll move response up a bit...

utils/crypto.cpp
rnistuk: Changed the previous throw to a LOG(error)

PR #137, opened 2018-10-03
KEP-378: High water mark, low water mark

isabelsavannah:


pbft/pbft.cpp
```diff
@@ -592,6 +588,9 @@ pbft::maybe_stabilize_checkpoint(const checkpoint_t& cp)
     this->clear_local_checkpoints_until(cp);
     this->clear_checkpoint_messages_until(cp);
     this->clear_operations_until(cp);
+
+    this->low_water_mark = std::max(this->low_water_mark, cp.first);
```
ebruck: shouldn't that be std::min?

isabelsavannah: No, both high and low water marks increase with checkpoints.

PR #138, opened 2018-10-03
KEP-486 Peer Can add itself to Peer List

rnistuk:
Peers can now add themselves to the swarm.


rnistuk:
Ooops, made a small change *after* the rebase with devel...

rnistuk:
Fixed a bug that macOS missed, rebased with devel and re submitted.

options/options.cpp
```diff
@@ -243,3 +243,10 @@ options::peer_validation_enabled() const
     //TODO: Remove this
     return this->raw_opts.get<bool>(PEER_VALIDATION_ENABLED);
 }
+
+
+std::string
+options::get_signed_key() const
```
ebruck: how about: get_signing_key?

utils/container.hpp
```diff
@@ -0,0 +1,42 @@
...
+#include <bootstrap/bootstrap_peers_base.hpp>
+#include <random>
+
+namespace
+{
+    std::mt19937 gen(std::time(0)); //Standard mersenne_twister_engine seeded with rd()
+}
+
+
+namespace bzn::utils::container
```
ebruck: Is this really going to be used elsewhere?

utils/container.hpp
```diff
@@ -0,0 +1,42 @@
...
+
+#pragma once
+
+#include <include/bluzelle.hpp>
+#include <bootstrap/bootstrap_peers_base.hpp>
+#include <random>
+
+namespace
+{
+    std::mt19937 gen(std::time(0)); //Standard mersenne_twister_engine seeded with rd()
```
ebruck: anywhere we include this it will be initialized? Not sure the value of the container.hpp.

utils/container.hpp
rnistuk: Yeah, I know. But I really wanted to unit test this one, so I just put in utils. My other thought was that it was a good start for a container library.

I'd like to keep it there, but I'm willing to move it to an anonymous namespace. I can live without the unit test now,.

utils/container.hpp
rnistuk: Yeah, that's a bad... I'll create a container.cpp, and break it up, for now. 

You know, what? I'll blow away the container.hpp and just move the function into the module where it's used..

PR #142, opened 2018-10-11
KEP-331: Cryptography for message authentication and integrity

isabelsavannah:


crypto/crypto.hpp
```diff
@@ -0,0 +1,52 @@
...
+#pragma once
+
+#include <crypto/crypto_base.hpp>
+#include <options/options_base.hpp>
+#include <proto/bluzelle.pb.h>
+#include <openssl/evp.h>
+
+namespace bzn
+{
+    class crypto : public crypto_base
```
ebruck: please prefix crypto_base with bzn

crypto/crypto.hpp
```diff
@@ -0,0 +1,52 @@
...
+    {
+    public:
+
+        crypto(std::shared_ptr<bzn::options_base> options);
+
+        void start();
```
ebruck: What's the reason behind a start? Can it not be done in the constructor?
Typically with use start because of enable_shared_from_this issues.

crypto/crypto.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+        bool sign(wrapped_bzn_msg& msg) override;
+
+        bool verify(const wrapped_bzn_msg& msg) override;
+
+        ~crypto() override;
```
ebruck: Please add virtual

crypto/crypto.cpp
```diff
@@ -0,0 +1,169 @@
...
+            && (key = EVP_PKEY_new())
+            && (1 == EVP_PKEY_set1_EC_KEY(key, pubkey))
+
+            // Perform the signature validation
+            && (context = EVP_MD_CTX_new())
+            && (1 == EVP_DigestVerifyInit(context, NULL, EVP_sha512(), NULL, key))
+            && (1 == EVP_DigestVerifyUpdate(context, msg.payload().c_str(), msg.payload().length()))
+            && (1 == EVP_DigestVerifyFinal(context, reinterpret_cast<const unsigned char*>(msg.signature().c_str()), msg.signature().length()));
+
+    if (context) EVP_MD_CTX_free(context);
```
ebruck: testing for null is not required.

crypto/crypto.cpp
```diff
@@ -0,0 +1,169 @@
...
+    if(this->options->get_simple_options().get<bool>(bzn::option_names::CRYPTO_ENABLED_OUTGOING))
+    {
+        this->load_private_key();
+    }
+}
+
+bool
+crypto::verify(const wrapped_bzn_msg& msg)
+{
+    BIO* bio = NULL;
```
ebruck: can we wrap these in std:unique_ptr?

crypto/crypto.cpp
```diff
@@ -0,0 +1,169 @@
...
+     * pollute our own logs with them (but we still have to clear the error state)
+     */
+    ERR_clear_error();
+
+
+    return result;
+}
+
+bool
+crypto::sign(wrapped_bzn_msg& msg)
```
ebruck: I like to add bzn:: at least in the function parameters

crypto/crypto.cpp
```diff
@@ -0,0 +1,169 @@
...
+        msg.set_sender(this->options->get_uuid());
+    }
+
+    if (msg.sender() != options->get_uuid())
+    {
+        LOG(error) << "Cannot sign message purportedly sent by " << msg.sender();
+        return false;
+    }
+
+    EVP_MD_CTX* context = NULL;
```
ebruck: use std::unique_ptr

crypto/crypto.cpp
```diff
@@ -0,0 +1,169 @@
...
+    this->log_openssl_errors();
+
+    return result;
+}
+
+bool
+crypto::load_private_key()
+{
+    auto filename = this->options->get_simple_options().get<std::string>(bzn::option_names::NODE_PRIVATEKEY_FILE);
+    ::FILE* fp;
```
ebruck: If we must use "C" style here you can still wrap this in a unique_ptr

crypto/crypto.cpp
```diff
@@ -0,0 +1,169 @@
...
+    while((last_error = ERR_get_error()))
+    {
+        ERR_error_string(last_error, buffer);
+        LOG(error) << buffer;
+    }
+}
+
+crypto::~crypto()
+{
+    if (this->private_key_EC) OPENSSL_free(this->private_key_EC);
```
ebruck: no need to test for null

crypto/crypto.hpp
isabelsavannah: Why? Shouldn't there be nothing that extends crypto?

crypto/crypto.cpp
isabelsavannah: wrapped_bzn_msg isn't actually in bzn; it's from protobuf

node/node.cpp
```diff
@@ -25,12 +25,14 @@ namespace
 
 
 node::node(std::shared_ptr<bzn::asio::io_context_base> io_context, std::shared_ptr<bzn::beast::websocket_base> websocket, std::shared_ptr<chaos_base> chaos, const std::chrono::milliseconds& ws_idle_timeout,
-    const boost::asio::ip::tcp::endpoint& ep)
+    const boost::asio::ip::tcp::endpoint& ep, std::shared_ptr<bzn::crypto_base> crypto, std::shared_ptr<bzn::options_base> options)
```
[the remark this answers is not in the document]
isabelsavannah: Conceptually it could change at runtime (key rotations), but my thinking was just that I'd rather not have the signature of the module have to change every time it requires a new option; you shouldn't have to directly handle all the options it wants to build a crypto. In particular, the chaos module already has six different options. It would be a pain to pass them all directly.

crypto/crypto.hpp
```diff
@@ -23,30 +23,31 @@
...
-
         bool sign(wrapped_bzn_msg& msg) override;
 
         bool verify(const wrapped_bzn_msg& msg) override;
 
-        ~crypto() override;
-
     private:
 
+        using EC_KEY_ptr = std::unique_ptr<EC_KEY, decltype(&::EC_KEY_free)>;
```
ebruck: typically we'd use:

EC_KEY_ptr_t

PR #144, opened 2018-10-18
KEP-574 pbft - configurations

paularchard:
Code for storing PBFT peers in an abstract "configuration" that can be versioned and passed as a message. The configuration class stores an individual config. The config_store class stores multiple configurations in order of adding, accesses them by hash, supports setting a configuration "enabled" and selecting a current configuration, as well as purging old ones.

The main PBFT code was updated to store the initial peers list in configuration and access it from there.

### DIFF
--- a/pbft/pbft_config_store.cpp
+++ b/pbft/pbft_config_store.cpp
@@ -21,7 +21,7 @@ pbft_config_store::add(pbft_configuration::shared_const_ptr config)
 {
     // TODO - should we be making a copy here instead?
     // currently the added config could be changed externally after being added
-    return (this->configs.insert(std::make_pair(this->next_index++, std::make_pair(config, false)))).second;
+    return (this->configs.insert(std::make_pair(this->next_index++, std::make_pair(std::move(config), false)))).second;
 }
 
 pbft_config_store::config_map::const_iterator
--- a/pbft/pbft_config_store.cpp
+++ b/pbft/pbft_config_store.cpp
@@ -1,0 +1,104 @@
+// Copyright (C) 2018 Bluzelle
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU Affero General Public License, version 3,
+// as published by the Free Software Foundation.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU Affero General Public License for more details.
+//
+// You should have received a copy of the GNU Affero General Public License
+// along with this program. If not, see <http://www.gnu.org/licenses/>.
+
+#include <pbft/pbft_config_store.hpp>
+
+using namespace bzn;
+
+pbft_config_store::pbft_config_store()
+    : current_index(0), next_index(1)
+{
+}
+
+bool
+pbft_config_store::add(pbft_configuration::shared_const_ptr config)
+{
+    // TODO - should we be making a copy here instead?
+    // currently the added config could be changed externally after being added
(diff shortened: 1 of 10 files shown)


ebruck:
Can you squash all commits into one for easier reviewing?

pbft/pbft_configuration.cpp
```diff
@@ -0,0 +1,214 @@
...
+    return json;
+}
+
+hash_t
+pbft_configuration::create_hash() const
+{
+    // TODO: better hash function
+
+    auto json = this->to_json().toStyledString();
+    size_t h = std::hash<std::string>{}(json);
```
ebruck: const size_t h or get rid of the temp variable.

ebruck: I'd even get rid of json.

ebruck: btw, I like one liners. :)

pbft/pbft_configuration.cpp
```diff
@@ -0,0 +1,214 @@
...
+    // and finally get each peer sequentially by sorted index and add to the destination vector
+    auto sorted = std::make_shared<std::vector<bzn::peer_address_t>>();
+    std::transform(indicies.begin(), indicies.end(), std::back_inserter(*sorted),
+        [&unordered_peers_list](auto& peer_index)
+        {
+           return unordered_peers_list[peer_index];
+        }
+    );
+
+    this->sorted_peers = sorted;
```
ebruck: std::move?

pbft/pbft_configuration.cpp
```diff
@@ -0,0 +1,214 @@
...
+    this->cached_hash = this->create_hash();
+}
+
+bool
+pbft_configuration::conflicting_peer_exists(const bzn::peer_address_t &peer) const
+{
+    for (auto p : this->peers)
```
ebruck: auto&

pbft/pbft_configuration.cpp
```diff
@@ -0,0 +1,214 @@
...
+    {
+        if (p.uuid == peer.uuid || p.name == peer.name)
```
ebruck: braces

pbft/pbft_configuration.cpp
```diff
@@ -0,0 +1,214 @@
...
+            return true;
+
+        if (p.host == peer.host)
+        {
+            if (p.port == peer.port || p.http_port == peer.http_port)
+                return true;
```
ebruck: braces

pbft/pbft_configuration.cpp
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    return false;
+}
+
+bool
+pbft_configuration::valid_peer(const bzn::peer_address_t &peer) const
+{
+    if (peer.name.empty() || peer.uuid.empty() || peer.host.empty() || !peer.port || !peer.http_port)
+        return false;
```
ebruck: braces

ebruck:
@isabelsavannah please also review this. I'm still coming up to speed on PBFT.

paularchard:
I've implemented and pushed the changes requested. The only exception was
the change to config_store::add to use std::move. I have tests where a
configuration that's been added is still used afterwards to check results.

Thanks,
Paul

pbft/pbft_config_store.cpp
```diff
@@ -0,0 +1,104 @@
...
+    : current_index(0), next_index(1)
+{
+}
+
+bool
+pbft_config_store::add(pbft_configuration::shared_const_ptr config)
+{
+    // TODO - should we be making a copy here instead?
+    // currently the added config could be changed externally after being added
+    return (this->configs.insert(std::make_pair(this->next_index++, std::make_pair(config, false)))).second;
```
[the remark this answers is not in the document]
paularchard: using std::move here would break tests that still use the config after adding it.

PR #151, opened 2018-10-25
KEP-573: dynamic peering

paularchard:
See https://bluzelle.atlassian.net/browse/KEP-573

Isabel has already seen this, apart from a few cosmetic changes.

### DIFF
--- a/pbft/pbft.cpp
+++ b/pbft/pbft.cpp
@@ -150,11 +150,6 @@ pbft::handle_bzn_message(const wrapped_bzn_msg& msg, std::shared_ptr<bzn::sessio
 void
 pbft::handle_membership_message(const wrapped_bzn_msg& msg, std::shared_ptr<bzn::session_base> /*session*/)
 {
-    if (msg.type() != BZN_MSG_PBFT_MEMBERSHIP)
-    {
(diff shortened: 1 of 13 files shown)


pbft/pbft.cpp
```diff
@@ -145,6 +147,36 @@ pbft::handle_bzn_message(const wrapped_bzn_msg& msg, std::shared_ptr<bzn::sessio
     this->handle_message(inner_msg, msg);
 }
 
+void
+pbft::handle_membership_message(const wrapped_bzn_msg& msg, std::shared_ptr<bzn::session_base> /*session*/)
+{
+    if (msg.type() != BZN_MSG_PBFT_MEMBERSHIP)
```
ebruck: I don't believe we need these tests. Node has tests for it's message delivery.

pbft/pbft.cpp
```diff
@@ -757,3 +866,58 @@ pbft::current_peers() const
...
+        return true;
+    }
+
+    return false;
+}
+
+bool
+pbft::proposed_config_is_acceptable(std::shared_ptr<pbft_configuration> config)
+{
+    (void) config;
```
ebruck: I don't like this. I would rather you comment out the param like so:

/* config */

PR #155, opened 2018-10-30
KEP-755 Bad Config (actually black listed uuid) Causes Uncaught Excep…

rnistuk:
…tion

### DIFF
--- a/bootstrap/bootstrap_peers.cpp
+++ b/bootstrap/bootstrap_peers.cpp
@@ -112,8 +112,7 @@ bootstrap_peers::initialize_peer_list(const Json::Value& root, bzn::peers_list_t
             // uuids
             if (bzn::utils::blacklist::is_blacklisted(uuid))
             {
-                LOG(warning) << "Ignoring blacklisted node with uuid: [" << uuid << "]";
-                continue;
+                throw std::runtime_error(std::string("This node has been actively disallowed from the Bluzelle network. Please contact support@bluzelle.com."));
             }
         }
 
--- a/bootstrap/bootstrap_peers.cpp
+++ b/bootstrap/bootstrap_peers.cpp
@@ -112,7 +112,8 @@ bootstrap_peers::initialize_peer_list(const Json::Value& root, bzn::peers_list_t
             // uuids
             if (bzn::utils::blacklist::is_blacklisted(uuid))
             {
-                throw std::runtime_error(std::string("This node has been actively disallowed from the Bluzelle network. Please contact support@bluzelle.com."));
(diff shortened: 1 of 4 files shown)


bootstrap/bootstrap_peers.cpp
```diff
@@ -112,8 +112,7 @@ bootstrap_peers::initialize_peer_list(const Json::Value& root, bzn::peers_list_t
             // uuids
             if (bzn::utils::blacklist::is_blacklisted(uuid))
             {
-                LOG(warning) << "Ignoring blacklisted node with uuid: [" << uuid << "]";
-                continue;
+                throw std::runtime_error(std::string("This node has been actively disallowed from the Bluzelle network. Please contact support@bluzelle.com."));
```
paularchard: This means that if any peer in the peers.json file is blacklisted we will throw and exit?  Is that the behavior we want?

raft/raft.cpp
```diff
@@ -1512,6 +1512,12 @@ raft::add_self_to_swarm()
     const auto this_address = std::find_if(all_peers.begin(), all_peers.end(),
             [&](const auto& peer){return peer.uuid == this->get_uuid();});
 
+    if (this_address == all_peers.end())
+    {
+        LOG(error) << "Unable to add this peer to the swarm. It may not be in the peers.json file, or it may be blacklisted.";
+        return;
```
paularchard: In this case we couldn't add ourselves to the swarm. Is that fatal? If so show we throw here?

bootstrap/bootstrap_peers.cpp
rnistuk: Wow, thanks for *that* catch! Fixed.

raft/raft.cpp
rnistuk: Yeah, I was struggling with that. It could simply be that the owner of the node did not add the node info to peers.json... I thought that that wasn't as bad a thing as a blacklisted peer..

But, now that I think about it, bailing right away seems like the nicest thing to do...

bootstrap/bootstrap_peers.cpp
rnistuk: Moved this test to raft.

raft/raft.cpp
rnistuk: Now the blacklist test is done in raft, in the case that the node listing was not in the default quorum.

rnistuk:
Thanks Paul, that is a good point. I have made a note in my "Skunkworks Refactoring Todo List" to change the "peers.json" to something more generic.

PR #163, opened 2018-11-06
KEP-769: Refactor wrapped_bzn_msg to bzn_envelope

isabelsavannah:


crypto/crypto.cpp
```diff
@@ -35,8 +35,24 @@ crypto::crypto(std::shared_ptr<bzn::options_base> options)
     }
 }
 
+const std::string&
+crypto::extract_payload(const bzn_envelope& msg)
+{
+    switch (msg.payload_case())
+    {
+        case bzn_envelope::kPbftRequest : return msg.pbft_request();
```
ebruck: This is not our coding standard. Each return should be contained withing a block below "case"

crypto/crypto.cpp
```diff
@@ -35,8 +35,24 @@ crypto::crypto(std::shared_ptr<bzn::options_base> options)
...
+        case bzn_envelope::kDatabaseResponse : return msg.database_response();
+        case bzn_envelope::kJson : return msg.json();
+        case bzn_envelope::kAudit : return msg.audit();
+        case bzn_envelope::kPbft : return msg.pbft();
+        case bzn_envelope::kPbftMembership : return msg.pbft_membership();
+        default : throw std::runtime_error(
```
ebruck: please reformat

crypto/crypto.hpp
```diff
@@ -46,6 +46,8 @@ namespace bzn
 
         void log_openssl_errors();
 
+        const std::string& extract_payload(const bzn_envelope&);
```
ebruck: missing param variable

proto/bluzelle.proto
```diff
@@ -18,31 +18,28 @@ import "database.proto";
...
 
+message bzn_envelope
+{
+    string sender = 1;
+    bytes signature = 2;
+
+    oneof payload
+    {
+        bytes pbft_request = 3;
+        bytes database_response = 4;
```
ebruck: Will a response ever arrive via this message type?

isabelsavannah: no, but it'll be sent out via it

PR #165, opened 2018-11-07
KEP-613: New node gets caught up from old nodes

paularchard:
Implementation of node state catchup. When a node sees a stable checkpoint that it doesn't have locally, it assumes it is behind and requests the database state at that point. When it receives the state it populates its database, which allows it to apply any pending operations after the checkpoint and get caught up.

### DIFF
--- a/pbft/pbft.cpp
+++ b/pbft/pbft.cpp
@@ -272,7 +272,6 @@ pbft::handle_request(const pbft_request& /*msg*/, const bzn::json_message& origi
     //TODO: conditionally discard based on timestamp - KEP-328
 
     //TODO: keep track of what requests we've seen based on timestamp and only send preprepares once - KEP-329
-
     auto op = setup_request_operation(original_msg.toStyledString(), session);
     this->do_preprepare(op);
 }
@@ -550,7 +549,7 @@ pbft::do_committed(const std::shared_ptr<pbft_operation>& op)
     }
 
     // TODO: this needs to be refactored to be service-agnostic
-    if (op->request.type() == PBFT_REQ_DATABASE)
+    if (op->get_request().type() == PBFT_REQ_DATABASE)
     {
         this->io_context->post(std::bind(&pbft_service_base::apply_operation, this->service, this->find_operation(op)));
     }
@@ -561,8 +560,12 @@ pbft::do_committed(const std::shared_ptr<pbft_operation>& op)
         msg->set_allocated_nullmsg(new database_nullmsg);
         pbft_request request;
         request.set_allocated_operation(msg);
-        auto new_op = std::make_shared<pbft_operation>(op->view, op->sequence, request, nullptr);
+        auto smsg = request.SerializeAsString();
+        auto new_op = std::make_shared<pbft_operation>(op->view, op->sequence
+            , this->crypto->hash(smsg), nullptr);
+        new_op->record_request(smsg);
         this->io_context->post(std::bind(&pbft_service_base::apply_operation, this->service, new_op));
+            request.SerializeAsString();
     }
 }
 
--- a/pbft/pbft.cpp
+++ b/pbft/pbft.cpp
@@ -156,7 +156,7 @@ pbft::handle_bzn_message(const bzn_envelope& msg, std::shared_ptr<bzn::session_b
 }
 
 void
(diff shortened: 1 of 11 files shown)


paularchard:
This commit also includes a significant amount of test fixture code intended to be used in cases where checkpoints need to be generated and such.

pbft/pbft.hpp
```diff
@@ -177,6 +184,8 @@ namespace bzn
         FRIEND_TEST(pbft_test, test_new_config_commit_handling);
         FRIEND_TEST(pbft_test, test_move_to_new_config);
 
+        friend class pbft_proto_test;
```
ebruck: you should not need this if you put the entire test suite in the bzn namespace.

pbft/pbft.cpp
```diff
@@ -369,6 +372,51 @@ pbft::handle_join_or_leave(const pbft_membership_msg& msg)
     }
 }
 
+void
+pbft::handle_get_state(const pbft_membership_msg& msg, std::shared_ptr<bzn::session_base> session) const
+{
+    // get stable checkpoint for request
+    checkpoint_t req_cp(msg.sequence(), msg.state_hash());
+
+    if (req_cp == this->latest_stable_checkpoint())
```
ebruck: I tend to like early returns instead of else cases.

pbft/pbft.cpp
```diff
@@ -474,8 +522,22 @@ pbft::do_committed(const std::shared_ptr<pbft_operation>& op)
         this->broadcast(this->wrap_message(msg));
     }
 
-    // Get a new shared pointer to the operation so that we can give pbft_service ownership on it
-    this->io_context->post(std::bind(&pbft_service_base::apply_operation, this->service, this->find_operation(op)));
+    // TODO: this needs to be refactored to be service-agnostic
+    if (op->request.type() == PBFT_REQ_DATABASE)
+    {
+        // Get a new shared pointer to the operation so that we can give pbft_service ownership on it
```
isabelsavannah: This comment was meant to be removed in one of my commits a while ago; I think it got accidentally readded by a rebase and it doesn't make sense anymore.

pbft/pbft.cpp
```diff
@@ -541,6 +603,16 @@ pbft::wrap_message(const pbft_msg& msg, const std::string& /*debug_info*/)
     return result.SerializeAsString();
 }
 
+bzn::encoded_message
+pbft::wrap_message(const pbft_membership_msg& msg, const std::string& /*debug_info*/) const
+{
```
isabelsavannah: this should return a bzn_envelope; node will not add a signature to an already-serialized message

pbft/pbft.cpp
```diff
@@ -369,6 +372,51 @@ pbft::handle_join_or_leave(const pbft_membership_msg& msg)
...
+            % msg.sequence(), msg.state_hash();
+    }
+}
+
+void
+pbft::handle_set_state(const pbft_membership_msg& msg)
+{
+    checkpoint_t cp(msg.sequence(), msg.state_hash());
+
+    if (this->unstable_checkpoint_proofs[cp].size() >= this->quorum_size() &&
```
isabelsavannah: Can you put an explanatory comment on this check? It took me several takes to understand its intent as "if we have checkpoint messages, but not the state, for the checkpoint"

pbft/pbft.cpp
```diff
@@ -650,24 +722,69 @@ pbft::unstable_checkpoints_count() const
...
+    pbft_membership_msg msg;
+    msg.set_type(PBFT_MMSG_GET_STATE);
+    msg.set_sequence(cp.first);
+    msg.set_state_hash(cp.second);
+
+    LOG(info) << boost::format("Requesting checkpoint state for hash %1% at seq %2%")
+        % cp.second % cp.first;
+
+    auto msg_ptr = std::make_shared<bzn::encoded_message>(this->wrap_message(msg));
+    this->node->send_message_str(make_endpoint(this->get_primary()), msg_ptr);
```
isabelsavannah: I'd rather not ask the primary. Instead:
- look at the checkpoint proof to determine who allegedly has the state (primary may not even have it)
- send the request to one of those nodes at random to spread load

PR #170, opened 2018-11-16
KEP-329: Primary should only send out preprepares once for each request

paularchard:
…even if the request is duplicated

Not sure if Ed needs to review this but added him out of courtesy.

### DIFF
--- a/pbft/pbft.cpp
+++ b/pbft/pbft.cpp
@@ -271,10 +271,11 @@ pbft::handle_request(const pbft_request& msg, const bzn::json_message& original_
         return;
     }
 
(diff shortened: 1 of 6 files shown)


pbft/pbft.cpp
```diff
@@ -269,10 +271,26 @@ pbft::handle_request(const pbft_request& /*msg*/, const bzn::json_message& origi
         return;
     }
 
-    //TODO: conditionally discard based on timestamp - KEP-328
+    if (msg.timestamp() < (this->now() - MAX_REQUEST_AGE_MS))
```
isabelsavannah: should also check if it is too far in the future

PR #172, opened 2018-11-20
KEP-809: Database requests no longer use json wrapper. Requests

isabelsavannah:
consistently handled, hashed as bzn_envelope. Breaks compatability with
old client version.

### DIFF
--- a/pbft/pbft.cpp
+++ b/pbft/pbft.cpp
@@ -247,10 +247,10 @@ pbft::preliminary_filter_msg(const pbft_msg& msg)
 }
 
 std::shared_ptr<pbft_operation>
-pbft::setup_request_operation(const bzn_envelope& request_env, const std::shared_ptr<session_base>& session)
+pbft::setup_request_operation(const bzn_envelope& request_env, const bzn::hash_t& request_hash, const std::shared_ptr<session_base>& session)
 {
     const uint64_t request_seq = this->next_issued_sequence_number++;
-    auto op = this->find_operation(this->view, request_seq, this->crypto->hash(request_env));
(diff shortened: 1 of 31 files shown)


ebruck:
@isabelsavannah what about the crud script?

pbft/pbft.cpp
```diff
@@ -262,43 +262,42 @@ pbft::setup_request_operation(const bzn::encoded_message& request, const request
...
         // TODO: send error message to client
-        LOG(info) << "Rejecting duplicate request: " << original_msg.toStyledString();
+        LOG(info) << "Rejecting duplicate request: " << request_env.ShortDebugString();
         return;
     }
-    this->saw_request(msg, hash);
+    this->saw_request(request_env, hash);
 
-    auto op = setup_request_operation(smsg, hash, session);
+    //TODO: keep track of what requests we've seen based on timestamp and only send preprepares once - KEP-329
```
paularchard: This comment was removed with KEP-329

pbft/pbft.cpp
```diff
@@ -262,43 +262,42 @@ pbft::setup_request_operation(const bzn::encoded_message& request, const request
...
+    auto op = setup_request_operation(request_env, session);
```
paularchard: I had previously moved the hashing out of setup_request_operation to avoid hashing the request twice. If you prefer it this way that's fine but I wasn't sure if crypto->hash() is an expensive operation and needed to be minimized.

isabelsavannah: fair enough; I had thought it was there only for the determinism concern.

PR #173, opened 2018-11-22
KEP-801: When forwarding a request to primary, replica holds session

isabelsavannah:
open and sends a response when the operation is eventually executed.

### DIFF
--- a/pbft/pbft.cpp
+++ b/pbft/pbft.cpp
@@ -14,6 +14,7 @@
 
 #include <pbft/pbft.hpp>
 #include <utils/make_endpoint.hpp>
+#include <utils/bytes_to_debug_string.hpp>
 #include <google/protobuf/text_format.h>
 #include <boost/beast/core/detail/base64.hpp>
 #include <boost/format.hpp>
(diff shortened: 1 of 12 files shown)


pbft/pbft.cpp
```diff
@@ -636,14 +666,22 @@ pbft::find_operation(uint64_t view, uint64_t sequence, const bzn::hash_t& req_ha
     auto lookup = operations.find(key);
     if (lookup == operations.end())
     {
-        LOG(debug) << "Creating operation for seq " << sequence << " view " << view << " req " << req_hash;
+        LOG(debug) << "Creating operation for seq " << sequence << " view " << view << " req " << bytes_to_debug_string(req_hash);
 
         std::shared_ptr<pbft_operation> op = std::make_shared<pbft_operation>(view, sequence, req_hash,
                 this->current_peers_ptr());
-        auto result = operations.emplace(std::piecewise_construct, std::forward_as_tuple(std::move(key)), std::forward_as_tuple(op));
+        auto emplace_result = operations.emplace(std::piecewise_construct, std::forward_as_tuple(std::move(key)), std::forward_as_tuple(op));
```
ebruck: I'm a big fan of structured bindings... I would do this as for readability

auto [it, added] = operation.emplace(...)

pbft/pbft.cpp
```diff
@@ -636,14 +666,22 @@ pbft::find_operation(uint64_t view, uint64_t sequence, const bzn::hash_t& req_ha
...
+        assert(emplace_result.second);
 
-        assert(result.second);
-        return result.first->second;
+        const auto session_pair = this->sessions_waiting_on_forwarded_requests.find(req_hash);
```
ebruck: same here

PR #176, opened 2018-11-26
Story/rnistuk/kep 332

rnistuk:


pbft/pbft.cpp
```diff
@@ -994,6 +1007,533 @@ pbft::get_high_water_mark()
...
+        minimum_sequence = pre_prepare.sequence() < minimum_sequence ? pre_prepare.sequence() : minimum_sequence;
+        maximum_sequence = pre_prepare.sequence() > maximum_sequence ? pre_prepare.sequence() : maximum_sequence;
+    }
+    return (maximum_sequence-minimum_sequence) == uint64_t(newview_msg.pre_prepare_messages_size());
+}
+
+bool
+pbft::is_valid_newview_message(const pbft_msg& msg, const bzn_envelope& /*original_msg*/) const
+{
+    // - does it contain 2f+1 viewchange messages
```
isabelsavannah: is there a reason we do this check in a different way from the others?

pbft/pbft.cpp
```diff
@@ -994,6 +1007,533 @@ pbft::get_high_water_mark()
...
+
+        viewchange_senders.insert(original_msg.sender());
+        std::set<std::pair<uint64_t, std::string>> sequence_request_pairs;
+
+        if (!this->get_sequences_and_request_hashes_from_proofs(viewchange_msg, sequence_request_pairs))
+        {
+            return false;
+        }
+
+        // each *pair* must match a pre-prepare in the the newview message...
```
isabelsavannah: My fault for forgetting to tell you this detail, but:
operations that come after the checkpoint which the new-view bases off of should be included.
So for example, if node A sends a view-change with checkpoint 100 and operations 102 and 105, but operation B sends a view change with checkpoint 200, then the new-view should use checkpoint 200, and should not have new preprepares for operations 102 and 105, because they're already included in checkpoint 200, and checkpoint 200 is proven valid.

pbft/pbft.cpp
```diff
@@ -994,6 +1007,533 @@ pbft::get_high_water_mark()
...
+        for (const auto sequence_requesthash : sequence_request_pairs)
+        {
+            uint64_t sequence{sequence_requesthash.first};
+            std::string request_hash{sequence_requesthash.second};
+
+            for (int j{0}; j <= msg.pre_prepare_messages_size(); ++j)
+            {
+                const bzn_envelope& pre_prepare_envelope = msg.pre_prepare_messages(j);
+                // TODO: do we need to cryptographically validate pre_prepare_envelope?
+                // TODO: do we need to check the sender?
```
isabelsavannah: yes, it should pass crypto->validate and it should have the same sender as the new-view.

pbft/pbft.cpp
```diff
@@ -994,6 +1007,533 @@ pbft::get_high_water_mark()
...
+                    return false;
+                }
+
+                // did we find the sequence?
+                if (sequence == pre_prepare.sequence()
+                    && request_hash == pre_prepare.request_hash())
+                {
+                    matched_pre_prepares.emplace_back(pre_prepare);
+                    break;
+                }
```
isabelsavannah: We also need to check that the preprepare has the same request_hash as the prepared_proof from the viewchange, and that its view is the view of the newview.

If the preprepare is not one from a viewchange, then it should be an empty request made to fill in a gap - we need to validate that its request_hash is the request_hash of an empty request.

isabelsavannah: If it would be a simpler implementation, it would be appropriate to use a magic number as the request hash of the empty request.

pbft/pbft.cpp
```diff
@@ -994,6 +1007,533 @@ pbft::get_high_water_mark()
...
+            }
+        }
+
+        if (matched_pre_prepares.size() != size_t(msg.pre_prepare_messages_size()))
```
isabelsavannah: I think this logic forgets the no-op requests?

pbft/pbft.cpp
```diff
@@ -994,6 +1007,533 @@ pbft::get_high_water_mark()
...
+pbft::fill_in_missing_pre_prepares(std::map<uint64_t, bzn_envelope> &pre_prepares)
+{
+    uint64_t last_sequence_number = pre_prepares.empty() ? 0 : (pre_prepares.end()--)->first;
+    for (uint64_t i = this->latest_stable_checkpoint().first + 1; i < last_sequence_number; ++i)
+    {
+        //  -- create a new preprepare for a no-op operation using this sequence number
+        if (pre_prepares.find(i) == pre_prepares.end())
+        {
+            auto msg = new database_msg;
+            msg->set_allocated_nullmsg(new database_nullmsg);
```
isabelsavannah: this is a database_msg, not a preprepare

pbft/pbft.cpp
```diff
@@ -994,6 +1007,533 @@ pbft::get_high_water_mark()
...
+ * @param msg
+ */
+void
+pbft::save_checkpoint(const pbft_msg& msg)
+{
+    pbft_msg message;
+    for (int i{0}; i < msg.checkpoint_messages_size(); ++i)
+    {
+        const auto checkpoint_msg = msg.checkpoint_messages(i);
+        bzn_envelope original_checkpoint;
```
isabelsavannah: the crypto->verify() check is necessary, the others are not.

pbft/pbft.cpp
```diff
@@ -994,6 +1007,533 @@ pbft::get_high_water_mark()
...
+            || !message.ParseFromString(original_checkpoint.pbft()))
+        {
+            continue;
+        }
+        this->handle_checkpoint(message, original_checkpoint);
+    }
+}
+
+void
+pbft::replica_broadcasts_viewchange(const pbft_msg& msg)
```
isabelsavannah: this name is inconsistent with others (nit)

pbft/pbft.cpp
```diff
@@ -994,6 +1007,533 @@ pbft::get_high_water_mark()
...
+    // replica sends VIEWCHANGE message
+    pbft_msg viewchange_message{
+            make_viewchange(
+                    msg.view()
+                    , this->latest_stable_checkpoint().first
+                    , this->stable_checkpoint_proof
+                    , this->prepared_operations_since_last_checkpoint())
+    };
+
+    this->view_is_valid = false;
```
isabelsavannah: don't we already set this in handle_failure?

pbft/pbft.cpp
```diff
@@ -994,6 +1007,533 @@ pbft::get_high_water_mark()
...
+    else
+    {
+        LOG(error) << "handle_viewchange - invalid viewchange message, ignoring";
+        return;
+    }
+
+    this->save_checkpoint(msg);
+
+    // When a replica recieves f + 1 view change messages it sends one as well
+    // should it be == or >=  ?
```
isabelsavannah: either way, the this->last_view_sent will prevent it from firing when n > f+1

pbft/pbft.cpp
```diff
@@ -994,6 +1007,533 @@ pbft::get_high_water_mark()
...
+                LOG (error) << "handle_viewchange - unable to parse viewchange message";
+                continue;
+            }
+
+            // Since sets require hashable objects we needed to use vectors and
+            // do our own uniqueness testing.
+            // TODO: use a set with a hash function
+            auto found_iter = std::find_if(std::cbegin(viewchange_messages) , std::cend(viewchange_messages) , [&](const auto& viewchange)
+            {
+                return !google::protobuf::util::MessageDifferencer::Equals(viewchange, pbft_viewchange);
```
isabelsavannah: I don't trust MessageDifferencer here, given the things we're doing with serialized messages.

pbft/pbft.cpp
```diff
@@ -994,6 +1007,533 @@ pbft::get_high_water_mark()
...
+    // When the primary of the new view recieves 2f valid view change messages for the new view
+    // the primary of new view broadcasts NEWVIEW
+
+    // TODO Refactor into method
+    {
+        // create the newview and and broadcast it
+
+        // viewchange->second is a set of bzn_envelope strings
+        // -- we want a of set<pbft_msg>
+
```
isabelsavannah: This is rather confusing. I suggest making pbft::valid_view_change_messages a
std::map<uint64_t, std::map<uuid_t, bzn_envelope>>- so a 2d map keyed by view and then by sender uuid.

pbft/pbft_operation.hpp
```diff
@@ -77,6 +77,18 @@ namespace bzn
 
         size_t faulty_nodes_bound() const;
 
+        const std::string& get_preprepare() const { return this->preprepare_message; };
```
isabelsavannah: these should return bzn_envelopes

pbft/pbft_operation.hpp
```diff
@@ -77,6 +77,18 @@ namespace bzn
...
+
+        const std::set<std::string> get_prepares() const
```
isabelsavannah: why is this implementation in the header?

pbft/pbft_operation.hpp
```diff
@@ -92,6 +104,11 @@ namespace bzn
         database_msg parsed_db;
         pbft_config_msg parsed_config;
 
+        std::string preprepare_message;
+        std::map<uuid_t, std::string> prepare_messages;  // uuid_t is the sender uuid, prepared messages
```
isabelsavannah: should be stored as bzn_envelope

pbft/pbft_operation.hpp
```diff
@@ -92,6 +104,11 @@ namespace bzn
...
+        bzn::encoded_message encoded_request;
+        pbft_request parsed_request;
```
isabelsavannah: these two fields have been removed

proto/pbft.proto
```diff
@@ -34,6 +35,24 @@ message pbft_msg
 
     // for checkpoints
     string state_hash = 6;
+
+    // for join/leave requests
+    pbft_peer_info peer_info = 7;
+
+    // for viewchange
+    // TODO replace string with wrapped_bzn_message
+    repeated string checkpoint_messages = 8;
```
isabelsavannah: bzn_envelope

proto/pbft.proto
```diff
@@ -34,6 +35,24 @@ message pbft_msg
...
+
+    // for viewchange, and newview (O)
+    // P, p_m, O
+    repeated prepared_proof prepared_proofs = 9;
+
+    // for newview
+    // V valid view change messages
+    repeated bytes viewchange_messages = 11;
```
isabelsavannah: bzn_envelope

proto/pbft.proto
```diff
@@ -89,4 +117,39 @@ message pbft_peer_info
...
     string uuid = 5;
+    pbft_request_type type = 6;
+    oneof msg
+    {
+        database_msg operation = 7;
+        pbft_internal_msg command = 8;
+    }
+    uint64 timestamp = 9;
+    string client = 10;
+}
```
isabelsavannah: this re-adds some stuff that has been removed. pbft_request no longer exists.

proto/pbft.proto
```diff
@@ -89,4 +117,39 @@ message pbft_peer_info
...
+    pbft_internal_msg_type type = 1;
+
+    // for new_config
+    string configuration = 2;
+}
+
+message prepared_proof
+{
+    bytes pre_prepare = 1;  // O
+    repeated bytes prepare = 2; // P, P_m
```
isabelsavannah: bzn_envelope

pbft/pbft.cpp
```diff
@@ -751,15 +764,15 @@ pbft::notify_audit_failure_detected()
 void
 pbft::handle_failure()
 {
-    LOG(fatal) << "Failure detected; view changes not yet implemented\n";
     this->notify_audit_failure_detected();
-    //TODO: KEP-332
```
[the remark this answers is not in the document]
rnistuk: Added 

LOG (error) << "handle_failure - PBFT failure - invalidating current view and sending VIEWCHANGE";

before this->notify_audit_failure_detected();

pbft/pbft.cpp
```diff
@@ -994,6 +1007,533 @@ pbft::get_high_water_mark()
...
+ * @param viewchange_message
+ * @return
+ */
+std::optional<bzn::checkpoint_t>
+pbft::validate_viewchange_checkpoints(const pbft_msg &viewchange_message) const
+{
+    //with viewchange_message.checkpoint_messages
+    //are there 2f+1 checkpoint messages each from a different nodes that are in the swarm
+    // that have the same checkpoint hash and valid signatures
+    size_t checkpoint_message_count = size_t(viewchange_message.checkpoint_messages_size());
```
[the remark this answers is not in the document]
rnistuk: Refactored to short circuit the two checks in this method.

pbft/pbft.cpp
```diff
@@ -994,6 +1007,533 @@ pbft::get_high_water_mark()
...
+    {
+        LOG(error) << "is_valid_viewchange_message - the checkpoint is invalid";
+        return false;
+    }
+
+    bzn_envelope pre_prepare_envelope;
+    // all the the prepared proofs are valid  (contains a pre prepare and 2f+1 mathcin prepares)
+    for (int i{0}; i < viewchange_message.prepared_proofs_size(); ++i)
+    {
+        if (!pre_prepare_envelope.ParseFromString(viewchange_message.prepared_proofs(i).pre_prepare()) || !this->is_peer(pre_prepare_envelope.sender()) || !this->crypto->verify(pre_prepare_envelope))
```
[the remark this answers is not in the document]
rnistuk: I will make a refactoring task for this.

pbft/pbft.cpp
rnistuk: Cool, here is the check:

if ((pre_prepare_envelope.sender() != original_msg.sender()) || !this->crypto->verify(pre_prepare_envelope))
                {
                    LOG (error) << "is_valid_newview_message - pre_prepare_envelope not valid";
                }

pbft/pbft.cpp
rnistuk: Yes, removed.

pbft/pbft.cpp
rnistuk: OK, removing the comment.

pbft/pbft_operation.hpp
rnistuk: Removed encoded_request and parsed_request.

proto/pbft.proto
rnistuk: Changed bytes to bzn_envelope

proto/pbft.proto
rnistuk: Removed pbft_request from pbft.proto

proto/pbft.proto
rnistuk: Changed to:

message prepared_proof
{
    bzn_envelope pre_prepare = 1;  // O
    repeated bzn_envelope prepare = 2; // P, P_m
}

proto/pbft.proto
rnistuk: Whups, missed this one. Changed string to ban_envelope... Fixing the fallout.

pbft/pbft.cpp
```diff
@@ -994,6 +1007,533 @@ pbft::get_high_water_mark()
...
+    for (int i{0}; i < newview_msg.pre_prepare_messages_size(); ++i)
+    {
+        const bzn_envelope pre_prepare_envelope{newview_msg.pre_prepare_messages(i)};
+        pbft_msg pre_prepare;
+        pre_prepare.ParseFromString(pre_prepare_envelope.pbft());
+
+        minimum_sequence = pre_prepare.sequence() < minimum_sequence ? pre_prepare.sequence() : minimum_sequence;
+        maximum_sequence = pre_prepare.sequence() > maximum_sequence ? pre_prepare.sequence() : maximum_sequence;
+    }
+    return (maximum_sequence-minimum_sequence) == uint64_t(newview_msg.pre_prepare_messages_size());
```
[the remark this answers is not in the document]
rnistuk: Re wrote pre_prepares_contiguous to walk through the pre prepares and check each sequence number with the previous one. This will catch missing, out of order and duplicate sequence numbers.

Empty pre prepares lists are considered to be contiguous.

pbft/pbft.cpp
rnistuk: In build new view, before collating pre prepare proofs, we ignore pre prepares with a sequence number lower than the highest checkpoint sequence number in the view change messages received from all of the nodes.

pbft/pbft.cpp
rnistuk: I  have refactored the code to remove the reliance on MessageDifferencer.

pbft/pbft_operation.hpp
rnistuk: Changed get_prepare to return bzn_envelope and propagated that change.

pbft/pbft.cpp
rnistuk: Paul helped me rewrite this section.

pbft/pbft.cpp
rnistuk: I've made this change and propagated it through the module.

PR #180, opened 2018-11-28
KEP-843, KEP-845: Update crud script to handle new pbft response style,

isabelsavannah:
fix some lingering bugs that prevented that response from being sent

the session is held as a shared_ptr instead of a weak_ptr because, between pbft's ack and crud's response, no operations are pending on that session and it will otherwise be cleaned up.

### DIFF
--- a/pbft/pbft_operation.cpp
+++ b/pbft/pbft_operation.cpp
@@ -193,13 +193,13 @@ pbft_operation::debug_string() const
 }
 
 void
-pbft_operation::set_session(std::weak_ptr<bzn::session_base> session)
+pbft_operation::set_session(std::shared_ptr<bzn::session_base> session)
 {
     this->listener_session = std::move(session);
     this->session_saved = true;
 }
 
-std::weak_ptr<bzn::session_base>
+std::shared_ptr<bzn::session_base>
 pbft_operation::session() const
 {
     return this->listener_session;
--- a/mocks/mock_session_base.hpp
+++ b/mocks/mock_session_base.hpp
@@ -35,5 +35,7 @@ namespace bzn {
             bzn::session_id());
         MOCK_METHOD0(close,
             void());
+        MOCK_CONST_METHOD0(is_open,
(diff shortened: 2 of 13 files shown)


pbft/pbft_operation.cpp
```diff
@@ -193,13 +193,13 @@ pbft_operation::debug_string() const
 }
 
 void
-pbft_operation::set_session(std::weak_ptr<bzn::session_base> session)
+pbft_operation::set_session(std::shared_ptr<bzn::session_base> session)
```
ebruck: We want weak pointers to not hold a reference to a connection that has gone away.

ebruck: There should always be a read scheduled.

isabelsavannah: There is no read scheduled between when pbft receives the database message and when the response is sent; send_message is what re-schedules the read. 

If we _do_ keep the session alive by immediately scheduling another read (this was previously happening as a side effect of pbft's early ack response), then when we try to send the actual response later, we crash on an assert inside boost with a comment saying "you can't have multiple pending async operations". We would have to intentionally schedule a (fake) read and the explicitly cancel it later (which session_base doesn't presently even allow), at which point it the read just feels like a nonobvious way to force the session to stay alive. The shared_ptr is owned by the pbft_operation anyway, which will automatically be cleaned based on the checkpoint process.

ebruck: Can you add a way to ask the session if the connection is still open? That way I can still bypass expensive db reads if no one is around to receive them.

ebruck: If you are getting a multiple read's scheduled then you don't have to do anything.

isabelsavannah: Maybe multiple reads work, but if you try to write while a read is outstanding it crashes on that assert.

"multiple reads scheduled" - reads on the database or on the socket?

PR #181, opened 2018-11-28
KEP-793: Refactor pbft_operation type to an abstract base, clean up its

isabelsavannah:
internals a bit to shrink the surface area of that interface. The plan here is to switch over to a new implementation of pbft_operation that stores its state in a storage_base instead of in memory.


pbft/pbft.hpp
```diff
@@ -86,6 +86,23 @@ namespace bzn
...
+        static size_t faulty_nodes_bound(size_t swarm_size);
+
+        /*
+         * minimum quorum size such that the majority of the quorum is guaranteed to be honest
+         * 2f+1
+         */
+        static size_t honest_majority_size(size_t swarm_size);
+
+        /*
+         * minimum quorum size such that at least one member is guarenteed to be honest
```
paularchard: typo: guaranteed

pbft/pbft.cpp
```diff
@@ -1185,3 +1185,21 @@ pbft::already_seen_request(const bzn_envelope& req, const request_hash_t& hash)
 
     return false;
 }
+
+size_t
+pbft::faulty_nodes_bound(size_t swarm_size)
+{
+    return (swarm_size-1)/3;
```
paularchard: spacing: (swarm_size - 1) / 3;

PR #184, opened 2018-11-30
KEP-747: Peer can add itself to peer list (PBFT Edition)

paularchard:


pbft/pbft.cpp
```diff
@@ -392,7 +398,7 @@ pbft::handle_commit(const pbft_msg& msg, const bzn_envelope& original_msg)
 }
 
 void
-pbft::handle_join_or_leave(const pbft_membership_msg& msg)
+pbft::handle_join_or_leave(const pbft_membership_msg& msg, std::shared_ptr<bzn::session_base> session)
 {
     if (!this->is_primary())
     {
```
isabelsavannah: was this copied from handle_request()? requests should be forwarded to the primary now

pbft/pbft.cpp
```diff
@@ -588,6 +624,24 @@ pbft::do_committed(const std::shared_ptr<pbft_operation>& op)
             // get rid of all other previous configs, except for currently active one
             this->configurations.remove_prior_to(config.get_hash());
         }
+
```
isabelsavannah: Should move this to after `op->advance_operation_stage(pbft_operation_stage::execute);`; it' has a relevant sanity check

paularchard: Will do. Side note though - seems like an easy attack for someone to send us a commit for an operation that isn't ready to commit and we'll throw an exception and exit...

paularchard: Actually I guess it's not that easy to get into that state.

isabelsavannah: Regardless of the likelihood of it happening here, you've got a point about that issue in general; I'm vaguely aware of several places where a bad peer could cause crashes by sending bad data. I'll make a backlog task to go over that.

PR #195, opened 2018-12-12
KEP-904: Caught up node doesn't switch to current view number

paularchard:
Another very small change


paularchard:
Modified code to only adopt a checkpoint if there is a consensus on which view to use.

pbft/pbft.cpp
```diff
@@ -530,6 +534,17 @@ pbft::handle_set_state(const pbft_membership_msg& msg)
 
         // TODO: validate the state data
         this->set_checkpoint_state(cp, msg.state_data());
+
+        if (msg.has_newview_msg())
+        {
+            pbft_msg newview;
```
isabelsavannah: crypto->verify the envelope

PR #197, opened 2018-12-13
KEP-900: viewchange validation fails due to unsigned messages

paularchard:


node/node.cpp
```diff
@@ -231,7 +231,7 @@ node::send_message(const boost::asio::ip::tcp::endpoint& ep, std::shared_ptr<bzn
         msg->set_sender(this->options->get_uuid());
     }
 
-    if(msg->signature().empty() && this->options->get_simple_options().get<bool>(bzn::option_names::CRYPTO_ENABLED_OUTGOING))
+    if (msg->signature().empty())
```
isabelsavannah: Should also remove the other corresponding check (node.cpp:150)

PR #201, opened 2018-12-13
KEP-902 Fixed a exception that occurs in is_valid_viewchange_message …

rnistuk:
…when no checkpoint has been set

### DIFF
--- a/pbft/pbft.cpp
+++ b/pbft/pbft.cpp
@@ -1122,7 +1122,7 @@ pbft::is_valid_viewchange_message(const pbft_msg& viewchange_message, const bzn_
     }
 
     auto valid_checkpoint_hashes = this->validate_and_extract_checkpoint_hashes(viewchange_message);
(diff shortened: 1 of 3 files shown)


pbft/pbft.cpp
```diff
@@ -1122,7 +1122,7 @@ pbft::is_valid_viewchange_message(const pbft_msg& viewchange_message, const bzn_
     }
 
     auto valid_checkpoint_hashes = this->validate_and_extract_checkpoint_hashes(viewchange_message);
-    if (valid_checkpoint_hashes.empty() && viewchange_message.sequence() != 0)
+    if (valid_checkpoint_hashes.empty() || viewchange_message.sequence() == 0)
```
paularchard: This check means that any viewchange where there hasn't been a checkpoint yet will fail.

rnistuk: I've re done the fix.

PR #202, opened 2018-12-15
Task/iscroggin/kep 794

isabelsavannah:


pbft/pbft.cpp
```diff
@@ -272,26 +278,32 @@ pbft::preliminary_filter_msg(const pbft_msg& msg)
...
 }
 
 void
 pbft::handle_request(const bzn_envelope& request_env, const std::shared_ptr<session_base>& session)
 {
+    const auto hash = this->crypto->hash(request_env);
+
+    if (session)
+    {
+        const auto hash = this->crypto->hash(request_env);
```
paularchard: Redundant

pbft/pbft.cpp
```diff
@@ -272,26 +278,32 @@ pbft::preliminary_filter_msg(const pbft_msg& msg)
...
+        if (this->sessions_waiting_on_forwarded_requests.find(hash) == this->sessions_waiting_on_forwarded_requests.end())
+        {
+            this->sessions_waiting_on_forwarded_requests[hash] = session;
+        }
+    }
+
```
paularchard: Could just do this->sessions_waiting_on_forwarded_requests.insert(std::make_pair(hash, session)); 

If element exists insert will fail.

PR #208, opened 2018-12-18
KEP-924 Removed "Uninteresting mock function call" warnings from pbft…

rnistuk:
Re did the submission by removing the NiceMock and adding the appropriate mock function calls.

This change is to unit tests only, and will be safe to merge into devel.

### DIFF
--- a/pbft/test/pbft_viewchange_test.cpp
+++ b/pbft/test/pbft_viewchange_test.cpp
@@ -30,7 +30,7 @@ namespace bzn
         std::shared_ptr<Mockcrypto_base>
         build_pft_with_mock_crypto()
         {
-            std::shared_ptr<Mockcrypto_base> mockcrypto = std::make_shared<NiceMock<Mockcrypto_base>>();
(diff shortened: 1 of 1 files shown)


pbft/test/pbft_viewchange_test.cpp
```diff
@@ -30,7 +30,7 @@ namespace bzn
         std::shared_ptr<Mockcrypto_base>
         build_pft_with_mock_crypto()
         {
-            std::shared_ptr<Mockcrypto_base> mockcrypto = std::make_shared<Mockcrypto_base>();
+            std::shared_ptr<Mockcrypto_base> mockcrypto = std::make_shared<NiceMock<Mockcrypto_base>>();
```
ebruck: are you 100% you want to squelch all calls?

rnistuk: Nope, changed my mind... Removed the Nice Mock and put in the appropriate expect calls.

PR #217, opened 2019-01-07
KEP-825 PBFT view change should propagate configuration change

paularchard:


pbft/pbft.cpp
```diff
@@ -710,6 +724,36 @@ pbft::do_committed(const std::shared_ptr<pbft_operation>& op)
...
+        LOG(error) << "handle_new_config_timeout error: " << ec.message();
+        return;
+    }
+
+    // send viewchange with new config
+    // TODO: should the current view be invalidated?
+    auto config = this->configurations.newest();
+    if (config)
+    {
+        this->view_is_valid = false;
```
isabelsavannah: I don't like duplicating this logic; can we invoke the method that the failure detector invokes? (perhaps renaming it)

pbft/pbft.cpp
```diff
@@ -1429,14 +1485,57 @@ pbft::handle_viewchange(const pbft_msg &msg, const bzn_envelope &original_msg)
...
+        {
+            LOG (debug) << "handle_newview - ignoring invalid NEWVIEW message while waiting to join swarm";
+            return;
+        }
+
+        // adopt checkpoint in newview message
+        pbft_msg viewchange;
+        viewchange.ParseFromString(msg.viewchange_messages(0).pbft());
+        this->save_checkpoint(viewchange);
+        this->in_swarm = swarm_status::joined;
```
isabelsavannah: Do we want to change our state if we still might reject the view change?

pbft/pbft.cpp
```diff
@@ -1610,6 +1712,64 @@ pbft::proposed_config_is_acceptable(std::shared_ptr<pbft_configuration> /* confi
     return true;
 }
 
+bool
+pbft::adopt_config_from_viewchange(const std::map<uuid_t, bzn_envelope>& viewchange_envelopes)
+{
+    std::map<bzn::hash_t, size_t> config_hashes;
```
isabelsavannah: I don't understand why we're deriving a config from the viewchange set. My understanding was that the new primary chooses a configuration for the newview it builds based on its own local state, and then replicas accept or reject it based on their local state.

paularchard: My thinking was that we should only use a configuration that is acceptable to enough peers, and that we have marked as acceptable ourselves (i.e. enabled).  The alternative would be to simply use our latest enabled configuration regardless. If that's your preference I can change it to do that instead.

pbft/pbft.cpp
paularchard: In theory this is benign - the node is not yet a member of the swarm and will not be until it receives a valid NEW_VIEW message. If you think it's worth it I can add code to reset to a null configuration in the case where the new_view is invalid.  Either way, there is a tautology here - the config contains the credentials to validate the new_view but if we don't trust the new_view we can't strictly use the config. I'll put in a comment to that effect and assume we will be retrieving config out-of-band in the future.

pbft/pbft.cpp

pbft/pbft.cpp
isabelsavannah: yeah, i guess now that you point it out, accepting the config is the scary part, and we do need a trust root for that anyway.

pbft/pbft.cpp
isabelsavannah: I'm just concerned about the case where the set of view_change messages does not agree on a configuration. In that case, the view change will fail, and the next time we try, presumably it would be with the same set of configurations? The original logic was designed (through of course it could have issues) such that the new primary would always choose a configuration that would be acceptable to at least 2f+1 replicas- and if we're doing it a different way we need to make sure that's maintained.

paularchard: Changed to always send latest prepared (active) configuration in new_view message

pbft/pbft.cpp
```diff
@@ -1610,6 +1707,23 @@ pbft::proposed_config_is_acceptable(std::shared_ptr<pbft_configuration> /* confi
     return true;
 }
 
+bool
+pbft::validate_config_in_newview(const pbft_msg& msg)
```
isabelsavannah: Is this still supposed to be here?

pbft/pbft.cpp
```diff
@@ -1571,7 +1665,7 @@ pbft::broadcast_new_configuration(pbft_configuration::shared_const_ptr config)
 }
 
 bool
-pbft::is_configuration_acceptable_in_new_view(hash_t config_hash)
+pbft::is_configuration_acceptable_in_new_view(const hash_t& config_hash)
 {
     return this->configurations.is_enabled(config_hash);
```
isabelsavannah: we should accept a configuration for a new view if we've accepted its preprepare; I think is_enabled corresponds to the operation being prepared, which is a stricter requirement.

pbft/pbft.cpp
paularchard: It's only used by a new node joining the swarm to do basic validation that the config is accepted by 2f+1 nodes. Again we can't do a full validation due to the trust chicken/egg situation though.

pbft/pbft.cpp
```diff
@@ -1596,16 +1684,15 @@ pbft::handle_config_message(const pbft_msg& msg, const std::shared_ptr<pbft_oper
...
-        this->configurations.set_current(config_hash);
-        this->configurations.remove_prior_to(config_hash);
+    if (this->configurations.current()->get_hash() == config_hash)
         return true;
-    }
 
-    return false;
+    assert(this->configurations.get(config_hash) != nullptr);
+    this->configurations.set_current(config_hash);
+    this->configurations.remove_prior_to(config_hash);
```
isabelsavannah: I think we shouldn't remove old configurations here (only in do_committed, where we are doing so)

PR #218, opened 2019-01-09
KEP-969: Send response to join request

paularchard:


paularchard:
Added a better test for new_config message content.

PR #222, opened 2019-01-17
KEP-975 Add snapshot support to rocksdb

ebruck:


storage/rocksdb_storage.cpp
```diff
@@ -256,20 +264,124 @@ rocksdb_storage::has_priv(const bzn::uuid_t& uuid, const  std::string& key)
...
 rocksdb_storage::create_snapshot()
 {
-    return false;
+    rocksdb::DumpOptions dump_options;
+
+    dump_options.db_path = this->db_path;
+    dump_options.dump_location = this->snapshot_file;
+    dump_options.anonymous = true;
+
+    std::shared_lock<std::shared_mutex> lock(this->lock); // lock for read access
```
rnistuk: Should the comment be "lock for write access"?

PR #225, opened 2019-01-22
fix for missing include file

paularchard:


paularchard:
Not sure why, but I can no longer reproduce the issue (undefined std::unordered_set when not including the header). I did an apt upgrade so that may have changed something. I'm closing this for now.

PR #228, opened 2019-01-23
 KEP-987: Refactor node to re-use sessions. remove raft. 

isabelsavannah:
The interesting files are

swarm.cpp
node.cpp
session.cpp
node_test_common.hpp
node_test.cpp
session_test.cpp

pretty much everything else is just collateral damage of the interface changes. I also want to make send_message operate by uuid instead of endpoint, but that requires some design thinking and this was too large already.


node/session.hpp
```diff
@@ -31,40 +31,56 @@ namespace bzn
...
+
         bzn::message_handler handler;
         bzn::protobuf_handler proto_handler;
 
-        std::mutex write_lock;
+        std::unique_ptr<bzn::asio::steady_timer_base> idle_timer;
+        const std::chrono::milliseconds ws_idle_timeout;
+
+        std::mutex socket_lock;
+        bool writing = false;
```
ebruck: std::atomic_bool?

node/node.cpp
```diff
@@ -24,50 +24,36 @@ namespace
...
 }
 
-
 void
 node::start()
 {
+    this->weak_priv_protobuf_handler =
+            [weak_self = weak_from_this()](auto msg, auto session)
+            {
+                auto strong_self = weak_self.lock();
```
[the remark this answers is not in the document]
isabelsavannah: to break the cyclical reference between session and node. It could be a shared pointer if node were changed to have shared ptrs.

node/node.cpp
```diff
@@ -165,63 +133,32 @@ node::priv_protobuf_handler(const bzn_envelope& msg, std::shared_ptr<bzn::sessio
...
+                    , this->weak_priv_protobuf_handler
+                    , this->options->get_ws_idle_timeout());
+            session->open_connection(this->websocket);
+            sessions.insert_or_assign(key, session);
+        }
 
-    if (this->chaos->is_message_dropped())
-    {
-        return;
+        session = this->sessions.at(key);
```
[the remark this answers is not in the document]
isabelsavannah: we made sure it was there at line 142, and we're holding the lock so that can't have changed.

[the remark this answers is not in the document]
ebruck: ah ok.. I see it now.

isabelsavannah:
The answer to "how do sessions get cleaned up" is lazily: when node tries to send a message on a session and discovers that it has closed, it replaces the session, which removes the last shared pointer. 

The case where it doesn't get cleaned up is where no message is ever sent to that endpoint (say it was a one-off client). This still doesn't leak a file descriptor (the socket times out and closes), but it does leak some memory. Changing it to a weak pointer makes it a bit cleaner, but it doesn't solve the issue: the weak pointer itself won't be removed from node's session map in this case, and it adds overhead on every message send. I think the solution with either kind of pointer is to regularly sweep the map and remove dead sessions; my intent was to delay implementing that because this PR is too large already.

(the other comments I agree with and I'll put a commit for them in later; I just wanted to unblock this conversation promptly)

ebruck:
> The answer to "how do sessions get cleaned up" is lazily: when node tries to send a message on a session and discovers that it has closed, it replaces the session, which removes the last shared pointer.
> 
> The case where it doesn't get cleaned up is where no message is ever sent to that endpoint (say it was a one-off client). This still doesn't leak a file descriptor (the socket times out and closes), but it does leak 

How? The member variable containing the websocket will never be destroyed, because session's destructor will never be called or until you replace the entry. A timeout does not close the FD as far as I understand networking. I  could be wrong. Have you tested this? Does Beast do this for you?


>some memory. Changing it to a weak pointer makes it a bit cleaner, but it doesn't solve the issue: the weak pointer itself won't be removed from node's session map in this case, and it adds overhead on 

I imagine any overhead would be nothing next to what it takes to sign the data or invoke any of the locks we are using. It's premature to worry about this and I'd rather let a profiler guide us instead.

>every message send. I think the solution with either kind of pointer is to regularly sweep the map and remove dead sessions; my intent was to delay implementing that because this PR is too large already.
> 

This is how subscription manager deals with it.

> (the other comments I agree with and I'll put a commit for them in later; I just wanted to unblock this conversation promptly)

Yeah, I'll try to finish my review ASAP.

node/node.cpp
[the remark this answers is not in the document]
ebruck: Unless you track the session in node, you will always create a new connection when pbft could of used an existing one.

node/session.cpp
```diff
@@ -15,182 +15,256 @@
...
 }
 
+void
+session::start_idle_timeout()
+{
+    this->activity = false;
 
-using namespace bzn;
+    this->idle_timer->expires_from_now(this->ws_idle_timeout);
+    this->idle_timer->async_wait(
```
ebruck: This is going to mess up pub/sub. I originally had this working, but the client never excepts the connection to close when it is waiting for subscribed data.

ebruck: Add a callback to node indicating that this session is going down instead.

node/session.hpp
```diff
@@ -31,40 +31,56 @@ namespace bzn
...
         bzn::session_id get_session_id() override { return this->session_id; }
 
         bool is_open() const override;
 
-    private:
-        void do_read();
+        void open_connection(std::shared_ptr<bzn::beast::websocket_base> ws_factory) override;
+        void accept_connection(std::shared_ptr<bzn::beast::websocket_stream_base> ws) override;
 
         void start_idle_timeout();
```
ebruck: make this private?

node/session_base.hpp
```diff
@@ -76,6 +53,16 @@ namespace bzn
          * @return id
          */
         virtual bzn::session_id get_session_id() = 0;
+
+        /**
+         * Create a new websocket connection for this session
+         */
+        virtual void open_connection(std::shared_ptr<bzn::beast::websocket_base> ws_factory) = 0;
```
ebruck: Please call it open instead.

node/session_base.hpp
```diff
@@ -76,6 +53,16 @@ namespace bzn
...
+
+        /**
+         * Accept an incoming connection on some websocket
+         */
+        virtual void accept_connection(std::shared_ptr<bzn::beast::websocket_stream_base> ws) = 0;
```
ebruck: Please call it accept instead.

node/session.cpp
```diff
@@ -15,182 +15,256 @@
...
+    this->writing = true;
 
-    std::lock_guard<std::mutex> lock(this->write_lock);
+    auto msg = this->write_queue.front();
+    this->write_queue.pop_front();
 
-    this->websocket->get_websocket().binary(true);
+    // so there will only be one instance of this callback
+    this->websocket->binary(true);
+    this->write_buffer = boost::asio::buffer(*msg);
```
ebruck: You're incurring an extra copy here.

isabelsavannah:
> I see that you've removed the ability to auto-close the session after a message is sent. I was wondering what the rationale for that is?

The sender of the message won't (and shouldn't) know if the message is sent over an existing session or a new one, and in the former case closing the session would potentially interfere with other messages. Ideally I'd like to have sessions be contained within node, so other stuff doesn't have to reason about them.

> Also, in the case of an error on session::write, you re-queue the message then close the socket. Is there any intention to try and restart communication? If not, what's the point in re-pushing the data (apart from logging a warning when the session is destroyed)?

Yes, my intent is for the session to be re-used with a new connection automatically.

node/node.cpp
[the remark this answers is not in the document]
isabelsavannah: I don't understand this? We do track the session in node.

node/node.cpp
```diff
@@ -165,63 +133,32 @@ node::priv_protobuf_handler(const bzn_envelope& msg, std::shared_ptr<bzn::sessio
...
-{
-    if (this->chaos->is_message_delayed())
+node::send_message_str(const boost::asio::ip::tcp::endpoint& ep, std::shared_ptr<bzn::encoded_message> msg) {
+    std::shared_ptr<bzn::session_base> session;
     {
-        const boost::asio::ip::tcp::endpoint ep_copy = ep;
-        this->chaos->reschedule_message(std::bind(&node::send_message_str, shared_from_this(), std::move(ep_copy), std::move(msg), close_session));
-        return;
-    }
+        std::lock_guard<std::mutex> lock(this->session_map_mutex);
```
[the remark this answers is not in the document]
isabelsavannah: My thinking is that the lock is never held for long (only looking up a session, or at worst initiating an async operation), so I didn't want to prematurely optimize

node/session.hpp
```diff
@@ -71,7 +71,7 @@ namespace bzn
         std::list<std::shared_ptr<bzn::encoded_message>> write_queue;
 
         bzn::protobuf_handler proto_handler;
-        bzn::session_death_handler death_handler;
+        bzn::session_shutdown_handler death_handler;
```
ebruck: sorry I should of also mentioned renaming the member variable as well.

PR #238, opened 2019-02-12
KEP-1092: Integration fixes for signature collation

isabelsavannah:
changed point of contact field to string
removed placeholder uuids from crud
moved message signing into session, changed method names to clarify
intent

### DIFF
--- a/node/node.cpp
+++ b/node/node.cpp
@@ -98,7 +98,8 @@ node::do_accept()
                         , self->chaos
                         , std::bind(&node::priv_protobuf_handler, self, std::placeholders::_1, std::placeholders::_2)
                         , self->options->get_ws_idle_timeout()
-                        , [](){});
+                        , [](){}
+                        , self->crypto);
 
                 session->accept(std::move(ws));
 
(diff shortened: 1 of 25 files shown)


node/session_base.hpp
```diff
@@ -37,6 +37,7 @@ namespace bzn
          * @param msg message
          */
         virtual void send_message(std::shared_ptr<bzn::encoded_message> msg) = 0;
+        virtual void send_signed_message(std::shared_ptr<bzn_envelope> msg) = 0;
```
ebruck: Please document the params and format this bode to be consistent .

node/node_base.hpp
```diff
@@ -57,7 +57,7 @@ namespace bzn
          */
         virtual void send_message_str(const boost::asio::ip::tcp::endpoint& ep, std::shared_ptr<bzn::encoded_message> msg) = 0;
 
-        virtual void send_message(const bzn::uuid_t &uuid, std::shared_ptr<bzn_envelope> msg) = 0;
+        virtual void send_signed_message(const bzn::uuid_t& uuid, std::shared_ptr<bzn_envelope> msg) = 0;
```
ebruck: reformat to be consistent and add comments

node/node.cpp
```diff
@@ -98,7 +98,8 @@ node::do_accept()
                         , self->chaos
                         , std::bind(&node::priv_protobuf_handler, self, std::placeholders::_1, std::placeholders::_2)
                         , self->options->get_ws_idle_timeout()
-                        , [](){});
+                        , [](){}
```
ebruck: Add a todo comment for this lambda?

isabelsavannah: That's the "on cleanup" callback, and no callback is required because node doesn't keep a reference to sessions that wrap incoming connections.

PR #239, opened 2019-02-12
KEP-1116: Daemon self-verifies own signature as a sanity check

isabelsavannah:


isabelsavannah:
That's... fair. I'd rather avoid mocking the interface of openssl, which seems to be what would be required there. crypto itself is already meant to be the encapsulation around that; adding a mock would add a whole extra layer. Short of that, I don't see how to straightforwardly get it to produce an invalid signature.

isabelsavannah:
I suppose I could move the self-validation logic into node, so that a mock_crypto could induce it, but that seems like worse code from every perspective except this unit test.

PR #244, opened 2019-02-21
kep-1031 - attempting to ensure that peers list is valid by throwing …

rnistuk:
…an runtime exception when a peer is found in the list that duplicates the current nodes' IP and port number

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,14 +101,14 @@ line argument:
 The configuration file is a JSON format file, as seen in the following example:
 
     {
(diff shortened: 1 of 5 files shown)


rnistuk:
If a user creates a peers config file that contains an entry that has the same IP and Port of the node current node, the node will post an error message that identifies that a peer in the list is hiding the current nodes IP address and node, and then exit via std::runtime_exception

rnistuk:
Thanks Ed!

PR #246, opened 2019-02-23
KEP-973: pbft variables reside in storage

paularchard:
Persistently stored variable framework and usage in PBFT.

I recommend reading the block comment in pbft_persistent_storage.hpp first. The intention was to hide all the complexity in there and make using the variables as easy as possible, but there are a couple of scenarios that are a little tricky.

### DIFF
--- a/pbft/pbft.hpp
+++ b/pbft/pbft.hpp
@@ -19,8 +19,10 @@
 #include <pbft/pbft_base.hpp>
 #include <pbft/pbft_failure_detector.hpp>
 #include <pbft/pbft_service_base.hpp>
+#include <pbft/pbft_persistent_state.hpp>
 #include <pbft/pbft_config_store.hpp>
 #include <pbft/operations/pbft_operation_manager.hpp>
+#include <storage/storage_base.hpp>
 #include <status/status_provider_base.hpp>
 #include <crypto/crypto_base.hpp>
 #include <proto/audit.pb.h>
@@ -41,6 +43,18 @@ namespace
     const double HIGH_WATER_INTERVAL_IN_CHECKPOINTS = 2.0; //TODO: KEP-574
     const uint64_t MAX_REQUEST_AGE_MS = 300000; // 5 minutes
     const std::string NOOP_REQUEST_HASH = "<no op request hash>";
+
+    std::string VIEW_KEY{"view"};
+    std::string NEXT_ISSUED_SEQUENCE_NUMBER_KEY{"next_issued_sequence_number"};
+    std::string ACCEPTED_PREPREPARES_KEY{"accepted_preprepares"};
+    std::string STABLE_CHECKPOINT_KEY{"stable_checkpoint"};
+    std::string LOCAL_UNSTABLE_CHECKPOINTS_KEY{"local_unstable_checkpoints"};
+    std::string UNSTABLE_CHECKPOINT_PROOFS_KEY{"unstable_checkpoint_proofs"};
+    std::string STABLE_CHECKPOINT_PROOF_KEY{"stable_checkpoint_proof"};
+    std::string VIEW_IS_VALID_KEY{"view_is_valid"};
+    std::string LAST_VIEW_SENT_KEY{"last_view_sent"};
+    std::string VALID_VIEWCHANGE_MESSAGES_FOR_VIEW_KEY{"valid_viewchange_messages_for_view"};
+    std::string SAVED_NEWVIEW_KEY{"saved_newview"};
 }
 
 
@@ -68,6 +82,7 @@ namespace bzn
             , std::shared_ptr<pbft_failure_detector_base> failure_detector
             , std::shared_ptr<bzn::crypto_base> crypto
             , std::shared_ptr<bzn::pbft_operation_manager> operation_manager
+            , std::shared_ptr<bzn::storage_base> storage
             );
 
         void start() override;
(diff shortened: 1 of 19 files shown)


pbft/pbft.hpp
```diff
@@ -41,6 +43,18 @@ namespace
     const double HIGH_WATER_INTERVAL_IN_CHECKPOINTS = 2.0; //TODO: KEP-574
     const uint64_t MAX_REQUEST_AGE_MS = 300000; // 5 minutes
     const std::string NOOP_REQUEST_HASH = "<no op request hash>";
+
```
ebruck: These should be const.

pbft/pbft.hpp
```diff
@@ -249,29 +267,29 @@ namespace bzn
         enum class swarm_status {not_joined, joining, waiting, joined};
         swarm_status in_swarm = swarm_status::not_joined;
```
isabelsavannah: shouldn't this be persisted? Or is the logic that we'll try to join and if we're already in it'll be handled appropriately

isabelsavannah:
What is the behaviour when there are multiple persist objects that "point" to the same key?

pbft/pbft.hpp
paularchard: done

pbft/pbft_persistent_state.hpp
```diff
@@ -0,0 +1,257 @@
...
+    // For non-collection types, declare an object of type persistent<type> and initialize it with
+    // { storage, default-value, unique-key }. If the variable already exists in storage it will be initialized
+    // with its stored state. If not it will use the provided default value.
+    //
+    // For collections, at instantiation time (e.g. insert, emplace, etc), insert the value as
+    // { storage, value, unique-object-name, key-name, key-name... }, where the key-names are the
+    // keys for each nested map, set, vector, etc.
+    // E.g. for map<int, persistent<std::string>> items, you would insert a string mapped to the integer 9 using:
+    // map.insert(std::make_pair(9, {storage, "value", "items", 9}); which would yield a storage key "items_00009".
+    //
```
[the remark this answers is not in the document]
paularchard: added info in block comment.

pbft/pbft_persistent_state.cpp
```diff
@@ -0,0 +1,253 @@
...
+using namespace bzn;
+
+std::string
+persist_base::escape(const std::string& input)
+{
+    std::string output;
+
+    for (auto ch : input)
+    {
+        output += (ch == SEPARATOR) ? std::string{SEPARATOR} + ch : std::string{ch};
```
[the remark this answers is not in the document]
paularchard: nice catch. will fix.

pbft/pbft.hpp
paularchard: There's currently no way to leave the swarm so once you join you're in, as far as the rest of the swarm is concerned. Re-joining is ignored. Perhaps it should be persisted but I can't think of a case where it would matter.

pbft/pbft_persistent_state.hpp
```diff
@@ -0,0 +1,257 @@
...
+    // In the case of nested maps, multiple subkeys are needed to guarantee the uniqueness of the generated
+    // storage key. e.g. for map<int, map<char, persistent<string>>> nesty; you would insert an item in the 7 map,
+    // mapped to the char 'x' like so: nesty[7]['x'] = {storage, "value", "nesty", 'x', 7}; which would yield a
+    // storage key "nesty_x_00007".
+    //
+    // For consistency, it's recommended that in nested collections the sub-keys be ordered from inner to outer.
+    //
```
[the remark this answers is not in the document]
paularchard: I'll put some thought into that.

paularchard:
> What is the behaviour when there are multiple persist objects that "point" to the same key?

In that case, the objects are "aliases", however modifying one instance will not modify the other but will update storage. There are cases where this is desirable (e.g. to destroy() an element in a set), but it needs to be handled with care.

I realize this has the potential for erroneous use. Trying to prevent it would require a fair amount of overhead but I'm open to suggestions.

pbft/pbft_persistent_state.hpp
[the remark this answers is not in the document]
paularchard: Added an internal registry for containers. Throws if you attempt to use a container without initializing first.

pbft/pbft_persistent_state.cpp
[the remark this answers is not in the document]
paularchard: Changed escaping to fix issue.

paularchard:
Added debug mode check to protect against unintentional aliasing. If a stored variable is altered through multiple aliases an exception will be thrown.

pbft/pbft.hpp
```diff
@@ -207,7 +223,7 @@ namespace bzn
 
         // VIEWCHANGE/NEWVIEW Helper methods
         void initiate_viewchange();
-        pbft_msg make_viewchange(uint64_t new_view, uint64_t n, const std::unordered_map<bzn::uuid_t, std::string>& stable_checkpoint_proof, const std::map<uint64_t, std::shared_ptr<bzn::pbft_operation>>& prepared_operations);
+        pbft_msg make_viewchange(uint64_t new_view, uint64_t n, const std::unordered_map<bzn::uuid_t, persistent<std::string>>& stable_checkpoint_proof, const std::map<uint64_t, std::shared_ptr<bzn::pbft_operation>>& prepared_operations);
```
isabelsavannah: const ref?

pbft/pbft_persistent_state.hpp
```diff
@@ -0,0 +1,334 @@
...
+        // initializing it with the provided default.
+        template <typename... K>
+        persistent(std::shared_ptr<bzn::storage_base> persist_storage, const T &default_value, const std::string& name, K... subkeys)
+            : storage(persist_storage), key(escape(name) + generate_key(subkeys...))
+        {
+            if (this->storage)
+            {
+                if (sizeof...(subkeys))
+                {
+                    if (initialized_containers.find(name) == initialized_containers.end())
```
isabelsavannah: I suppose they could still be using an uninitialized map after a different alias of the map was initialized somewhere? I guess we're not likely to get it perfectly, though.

PR #248, opened 2019-02-28
Story/rnistuk/kep 808

rnistuk:


pbft/pbft.cpp
```diff
@@ -1887,3 +1888,20 @@ uint32_t pbft::generate_random_number(uint32_t min, uint32_t max)
...
+void pbft::add_session_to_pool(const std::string& msg_hash, std::shared_ptr<bzn::session_base> session)
+{
+    if (session)
+    {
+        this->sessions_waiting_on_forwarded_requests[msg_hash] = session;
+        session->add_shutdown_handler([msg_hash, this]()
+                                        {
+                                            std::lock_guard<std::mutex> lock(this->pbft_lock);
+                                            auto it = this->sessions_waiting_on_forwarded_requests.find(msg_hash);
+                                            if (it != this->sessions_waiting_on_forwarded_requests.end() && !it->second->is_open())
```
isabelsavannah: Instead of checking that it's still open, check that it's the same session. Sessions already have unique ids, so that's easy. I'm concerned about a race condition where the old session timed out and we're already in the process of opening a new one, so the new one isn't open yet.

pbft/pbft.cpp
```diff
@@ -1887,3 +1888,20 @@ uint32_t pbft::generate_random_number(uint32_t min, uint32_t max)
     std::uniform_int_distribution<uint32_t> dist(min, max);
     return dist(gen);
 }
+
+void pbft::add_session_to_pool(const std::string& msg_hash, std::shared_ptr<bzn::session_base> session)
```
isabelsavannah: This name feels strange to me; it's not a pool of interchangeable sessions

pbft/pbft.cpp
rnistuk: I have replaced the !it->second->is_open() check with 
(it->second->get_session_id() == session->get_session_id())

pbft/pbft.cpp
rnistuk: renamed to add_session_to_sessions_waiting

rnistuk:
Thanks Isabel!

PR #251, opened 2019-03-01
kep-538 pbft_failure_detector should forget comnpleted request hashes…

rnistuk:
… some time after they become irrelevant

### DIFF
--- a/pbft/pbft_failure_detector.cpp
+++ b/pbft/pbft_failure_detector.cpp
@@ -88,8 +88,7 @@ pbft_failure_detector::request_executed(const bzn::hash_t& req_hash)
     std::lock_guard<std::mutex> lock(this->lock);
 
     this->outstanding_requests.erase(req_hash);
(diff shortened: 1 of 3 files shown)


pbft/pbft_failure_detector.cpp
```diff
@@ -98,4 +97,18 @@ pbft_failure_detector::register_failure_handler(std::function<void()> handler)
...
     this->failure_handler = handler;
+}
+
+void
+pbft_failure_detector::add_completed_request_hash(const bzn::hash_t& request_hash)
+{
+    this->completed_requests.emplace(request_hash);
+    this->completed_request_queue.push(request_hash);
+
+    if (max_completed_requests_memory < this->completed_requests.size())
```
isabelsavannah: I prefer while to if here; it shouldn't ever be different but if it is, then removing everything over the limit is the correct behavior. Suppose in the future we change the options module to support modifying options at runtime, for xample.

PR #256, opened 2019-03-07
KEP-1222: Collate monitor output, to reduce log spam and network

isabelsavannah:
traffic

### DIFF
--- a/options/simple_options.hpp
+++ b/options/simple_options.hpp
@@ -38,8 +38,8 @@ namespace bzn::option_names
     const std::string MEM_STORAGE = "mem_storage";
     const std::string MONITOR_ADDRESS = "monitor_address";
     const std::string MONITOR_PORT = "monitor_port";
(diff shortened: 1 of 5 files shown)


options/simple_options.hpp
```diff
@@ -38,6 +38,8 @@ namespace bzn::option_names
     const std::string MEM_STORAGE = "mem_storage";
     const std::string MONITOR_ADDRESS = "monitor_address";
     const std::string MONITOR_PORT = "monitor_port";
+    const std::string MONITOR_COLLATE = "MONITOR_COLLATE";
```
rnistuk: perhaps the text for 
	    const std::string MONITOR_COLLATE = "MONITOR_COLLATE";
	    const std::string MONITOR_COLLATE_INTERVAL_SECONDS = "MONITOR_COLLATE_INTERVAL_SECONDS";

should be lowercase.

PR #264, opened 2019-03-14
KEP-1243 Querying the keys of a non-existent database should return a…

ebruck:
…n error


rnistuk:
Cool! My keys script returns the correct error message now. Approving this pull request.

PR #276, opened 2019-03-20
KEP-1254: add detail to swarmdb --version

isabelsavannah:


isabelsavannah:
Can someone please double check that this doesn't break anything on mac? I believe these macros should exist on every platform we care about, but I can't really test that myself...

PR #277, opened 2019-03-20
kep-1165 create_db and update_db can only be used by Bluzelle

rnistuk:


storage/storage_base.hpp
```diff
@@ -49,7 +51,9 @@ namespace bzn
         {storage_result::db_not_found,    "DATABASE_NOT_FOUND"},
         {storage_result::db_exists,       "DATABASE_EXISTS"},
         {storage_result::access_denied,   "ACCESS_DENIED"},
-        {storage_result::delete_pending,  "DELETE_PENDING"}};
+        {storage_result::delete_pending,  "DELETE_PENDING"},
+        {storage_result ::db_create_denied, "DB_CREATE_DENIED"},
```
ebruck: We don't need more types. ACCESS_DENIED is good enough.

options/simple_options.hpp
```diff
@@ -47,6 +47,7 @@ namespace bzn::option_names
     const std::string WS_IDLE_TIMEOUT = "ws_idle_timeout";
     const std::string PEER_VALIDATION_ENABLED = "peer_validation_enabled";
     const std::string SIGNED_KEY = "signed_key";
+    const std::string BLUZELLE_PUBLIC_KEY = "bluzelle_public_key";
```
ebruck: This should be something like owner_public_key.

options/options_base.hpp
```diff
@@ -175,5 +175,12 @@ namespace bzn
          * @return string containing the signature
          */
         virtual std::string get_signed_key() const = 0;
+
+
+        /**
+         * Retrieve the path to the Bluzelle public key pem file
+         * @return string containing the path to the Bluzelle publi key pem file
+         */
+        virtual std::string get_bluzelle_public_key() const = 0;
```
ebruck: change to owner_public_key

options/simple_options.cpp
```diff
@@ -141,7 +141,11 @@ simple_options::build_options()
                         "require signed key for new peers to join swarm")
                 (SIGNED_KEY.c_str(),
                         po::value<std::string>(),
-                        "signed key for node's uuid");
+                        "signed key for node's uuid")
+                (BLUZELLE_PUBLIC_KEY.c_str(),
```
ebruck: Weren't we going to NOT load from disc?

options/options_base.hpp
ebruck: One way is enough. lets only use the conf

crud/crud.cpp
```diff
@@ -535,11 +535,16 @@ crud::handle_unsubscribe(const bzn::caller_id_t& /*caller_id*/, const database_m
...
-    bzn::storage_result result;
+    bzn::storage_result result{bzn::storage_result::ok};
 
     std::lock_guard<std::shared_mutex> lock(this->lock); // lock for write access
 
-    if (this->storage->has(PERMISSION_UUID, request.header().db_uuid()))
+    // if node is undefined here, there will be  segfault, when unit testing, make sure node is defined.
+    if (!this->node->bluzelle_public_key().empty() && (this->node->bluzelle_public_key() != caller_id))
+    {
+        result = bzn::storage_result::db_create_denied;
```
ebruck: Lets stick with access_denied.

storage/storage_base.hpp
rnistuk: OK, removing DB_CREATE_DENIED and DB_DELETE_DENIED, using ACCESS_DENIED instead.

options/simple_options.hpp
rnistuk: Changed to owner_public_key and propagated the changes.

options/options_base.hpp
rnistuk: Yeah, it only uses the conf.

options/simple_options.cpp
rnistuk: We aren't, the public key is fead from the conf.

crud/crud.cpp
rnistuk: Done

rnistuk: Removed the necessity for the node::owner_public_key method by adding an owner_public_key bzn::key_t member to crud - propogated this change to the rest of the code

options/simple_options.cpp
```diff
@@ -141,7 +141,11 @@ simple_options::build_options()
                         "require signed key for new peers to join swarm")
                 (SIGNED_KEY.c_str(),
                         po::value<std::string>(),
-                        "signed key for node's uuid");
+                        "signed key for node's uuid")
+                (OWNER_PUBLIC_KEY.c_str(),
+                        po::value<std::string>(),
+                        "path to the Bluzelle public key pem file");
```
ebruck: help text is wrong

crud/crud.hpp
```diff
@@ -93,6 +93,7 @@ namespace bzn
 
         std::once_flag start_once;
         std::shared_mutex lock; // for multi-reader and single writer access
+        bzn::key_t  owner_public_key;
```
ebruck: make this const and use std::move if you can in the constructor

options/simple_options.cpp
rnistuk: Changed to "swarm owner's public key"

PR #286, opened 2019-03-26
KEP-1200 API support for asking for the space left in a namespace

ebruck:


storage/mem_storage.cpp
```diff
@@ -137,7 +137,7 @@ mem_storage::remove(const bzn::uuid_t& uuid, const std::string& key)
         return bzn::storage_result::not_found;
     }
 
-    search->second.first -= record->second.size();
+    search->second.first -= (record->second.size() + key.size());
```
rnistuk: Whups, adding key size to the total size.. Good catch

crud/crud.cpp
```diff
@@ -490,13 +490,27 @@ crud::handle_size(const bzn::caller_id_t& /*caller_id*/, const database_msg& req
...
+    }
+
     const auto [keys, size] = this->storage->get_size(request.header().db_uuid());
 
     database_response response;
 
     response.mutable_size()->set_keys(keys);
     response.mutable_size()->set_bytes(size);
 
+    if (const auto max_size = boost::lexical_cast<uint64_t>(perms[MAX_SIZE_KEY]); max_size)
```
rnistuk: Yeah, if the max_size has been set, go ahead and return the size remaining

PR #291, opened 2019-03-27
KEP-1294: missing lock causing instability of swarm (devel)

paularchard:
Added a lock to pbft::handle_database_request
Cleaned up a couple of log messages to work better for large db values

### DIFF
--- a/pbft/pbft.cpp
+++ b/pbft/pbft.cpp
@@ -270,7 +270,7 @@ pbft::preliminary_filter_msg(const pbft_msg& msg)
 
         if (msg.sequence() <= this->low_water_mark)
         {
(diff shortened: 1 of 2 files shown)


pbft/pbft.cpp
```diff
@@ -1101,6 +1101,8 @@ pbft::handle_database_message(const bzn_envelope& msg, std::shared_ptr<bzn::sess
 {
     // TODO: timestamp should be set by the client. setting it here breaks the signature (which is correct).
 
+    std::lock_guard<std::mutex> lock(this->pbft_lock);
```
ebruck: Move this inside the if block. This removes the gains we made with switching to a shared lock when qreads are executed.

paularchard: Done

PR #295, opened 2019-04-01
KEP-1150 Swarm will reject any writes that exceed the hard limit size of the database

ebruck:


crud/crud.cpp
```diff
@@ -1068,3 +1083,37 @@ crud::flush_expiration_entries(const bzn::uuid_t& uuid)
...
+            return (size + request.create().key().size() + request.create().value().size() > max_size);
+        }
+
+        // test update
+        if (request.msg_case() == database_msg::kUpdate)
+        {
+            const auto prev_kv_size = this->storage->get_key_size(request.header().db_uuid(),
+                request.update().key());
+
+            if (prev_kv_size)
```
rnistuk: yeah, prev_kv_size is actually an option, so it could be null.

crud/crud.cpp
```diff
@@ -1068,3 +1083,37 @@ crud::flush_expiration_entries(const bzn::uuid_t& uuid)
...
+            {
+                return (size - *prev_kv_size + request.update().key().size() + request.update().value().size() > max_size);
+            }
```
rnistuk: I sort of want there to be an else in this case, with the same return value as a create would give... But I realize we would never get here if there was no previous key anyway...

storage/mem_storage.cpp
```diff
@@ -23,9 +23,9 @@ using namespace bzn;
 
 
 bzn::storage_result
-mem_storage::create(const bzn::uuid_t& uuid, const std::string& key, const std::string& value)
+mem_storage::create(const bzn::uuid_t& uuid, const bzn::key_t& key, const bzn::value_t& value)
```
rnistuk: oh.. bzn::key_t and bzn::value_t types... OK, I will use those from now on..

storage/mem_storage.cpp
```diff
@@ -23,9 +23,9 @@ using namespace bzn;
...
 {
-    std::lock_guard<std::shared_mutex> lock(this->lock); // lock for write access
+    std::lock_guard<std::shared_mutex> lock(this->kv_store_lock); // lock for write access
```
rnistuk: Nice, this gets rid of the "shadow" warnings. I'll do the same in crud.

PR #300, opened 2019-04-12
kep-1144 Add Swarm ID to all swarm<>swarm and swarm<>client messages

rnistuk:
Added swarm id to the bzn_envelope protobuf structure and propagated the changes through the code. The swarm id parameter is set in pbft::wrap_message method.


isabelsavannah:
I think this field should be checked and set in node, not pbft, right? There are a lot of things besides pbft that send messages.

rnistuk:
> I think this field should be checked and set in node, not pbft, right? There are a lot of things besides pbft that send messages.

Cool! On it. Thanks for the feedback

PR #307, opened 2019-04-26
kep-1358 Added swarm-id option to the crud script, updated the readme…

rnistuk:
… to reflect the addition of the swarm id argument


ebruck:
please use KEP instead of kep in the commit message

PR #309, opened 2019-05-01
KEP-1362 Updated darwin to 1.70

ebruck:
* update cmake/boost.cmake

### DIFF
--- a/cmake/boost.cmake
+++ b/cmake/boost.cmake
@@ -15,8 +15,8 @@
 include(ExternalProject)
 include(ProcessorCount)
 
-set(REQUIRED_BOOST "1.68.0")
-set(BOOST_URL_HASH "da3411ea45622579d419bfda66f45cd0f8c32a181d84adfa936f5688388995cf")
+set(REQUIRED_BOOST "1.70.0")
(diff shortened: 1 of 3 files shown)


cmake/boost.cmake
```diff
@@ -15,8 +15,8 @@
 include(ExternalProject)
 include(ProcessorCount)
 
-set(REQUIRED_BOOST "1.68.0")
-set(BOOST_URL_HASH "da3411ea45622579d419bfda66f45cd0f8c32a181d84adfa936f5688388995cf")
+set(REQUIRED_BOOST "1.70.0")
+set(BOOST_URL_HASH "882b48708d211a5f48e60b0124cf5863c1534cd544ecd0664bb534a4b5d506e9")
```
rnistuk: Updated required boost, and the hash is correct. Cool.

monitor/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ add_library(monitor
         monitor.cpp
         )
 
-add_dependencies(monitor boost)
+add_dependencies(monitor boost jsoncpp)
```
rnistuk: We forgot jsoncpp?

PR #311, opened 2019-05-03
Task/rnistuk/kep 1366

rnistuk:
Cleaned up the cmake files to ensure that the correct boost 1.70.0  include files are included, as opposed to those in the system folder. Also cleaned up a deprecated function warning.


CMakeLists.txt
```diff
@@ -63,7 +55,7 @@ find_package(CURL REQUIRED)
 
 include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 include_directories(SYSTEM ${OPENSSL_INCLUDE_DIR})
```
ebruck: remove these

CMakeLists.txt
```diff
@@ -63,7 +55,7 @@ find_package(CURL REQUIRED)
...
-include_directories(SYSTEM ${Boost_INCLUDE_DIRS})
+#include_directories(SYSTEM ${Boost_INCLUDE_DIRS})
```
ebruck: remove this

rnistuk: Whups, missed this. Done

CMakeLists.txt
rnistuk: Done

PR #317, opened 2019-05-15
KEP-1420 Added and corrected missing build steps for Ubuntu

nickroeker:


README.md
```diff
@@ -41,18 +41,31 @@ $ brew install ccache
...
+
+```text
+$ sudo apt-get install curl libcurl4-openssl-dev
+$ sudo apt-get install cmake
+```
+If your system packages don't have a new enough version, you can install a
+different CMake into `~/mycmake/` to avoid overwriting your system's `cmake`.
 
 ```text
+$ sudo apt-get install curl libcurl4-openssl-dev
```
ebruck: curl comes automatically with the dev package?

nickroeker: It does not unfortunately, assuming you mean `libcurl4-openssl-dev`. At least not in 18.04's default packages? I had to separately install it in a fresh VM after doing all other setup.

ebruck: strange (pulled from 19.04):

Package: libcurl4-openssl-dev
Source: curl
Version: 7.64.0-2ubuntu1
Architecture: amd64
Maintainer: Ubuntu Developers <ubuntu-devel-discuss@lists.ubuntu.com>
Installed-Size: 1472
**Depends: libcurl4 (= 7.64.0-2ubuntu1)**

nickroeker: You're correct of course, I must've goofed it somehow in the first round. `curl` itself seems to only be required by the instructions as well, as the pulled-in libcurl does just fine. Will update.

PR #326, opened 2019-06-10
KEP-1500 - added info to exception, also added cert directory for Ubuntu

paularchard:
debugging info to track down which variable is the problem.
Also added certificate directory for Ubuntu

### DIFF
--- a/utils/http_req.cpp
+++ b/utils/http_req.cpp
@@ -34,6 +34,7 @@ namespace
         "/system/etc/security/cacerts", // Android
         "/usr/local/share/certs",       // FreeBSD
         "/etc/pki/tls/certs",           // Fedora/RHEL
+        "/etc/ssl/certs",               // Ubuntu
         "/etc/openssl/certs",           // NetBSD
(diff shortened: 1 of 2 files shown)


utils/http_req.cpp
```diff
@@ -34,6 +34,7 @@ namespace
         "/system/etc/security/cacerts", // Android
         "/usr/local/share/certs",       // FreeBSD
         "/etc/pki/tls/certs",           // Fedora/RHEL
+        "/etc/ssl/certs",               // Ubuntu
```
ebruck: This really isn't needed as we do this and /cert is appended automatically:

# platform detection
if (APPLE)
    set(OPENSSL_BUILD_PLATFORM darwin64-x86_64-cc)
    set(OPENSSLDIR /usr/local/etc/openssl)
else()
    set(OPENSSL_BUILD_PLATFORM linux-generic64)
    set(OPENSSLDIR /etc/ssl)
endif()

ebruck: I guess you're not yet using our git lfs changes.

PR #327, opened 2019-06-10
KEP-1149 Swarm level API to determine the amount of space remaining

ebruck:


crud/crud.cpp
```diff
@@ -1201,22 +1246,46 @@ crud::flush_expiration_entries(const bzn::uuid_t& uuid)
 bool
 crud::operation_exceeds_available_space(const database_msg& request, const Json::Value& perms)
 {
+    const auto request_type = request.msg_case();
     const auto max_size = this->max_database_size(perms);
 
+    if (request_type == database_msg::kCreateDb || request_type == database_msg::kUpdateDb)
+    {
+        if (!this->max_swarm_storage)
```
rnistuk: ok, if max_swarm_storage is 0, then we ignore restrictions

crud/crud.cpp
```diff
@@ -1273,3 +1375,24 @@ crud::random_cache_replacement(const database_msg& request, size_t key_value_siz
         }
     }
 }
+
+
+std::string
+crud::get_name()
```
rnistuk: For use in the status module. cool

PR #331, opened 2019-06-17
KEP-823 Remove Full Request from Saved and Retransmitted Pre-prepares

rnistuk:


node/test/session_test.cpp
```diff
@@ -194,7 +194,7 @@ namespace bzn
         io2->shutdown();
     }
 
-    TEST_F(session_test2, additional_shutdown_handlers_can_be_added_to_session)
+    TEST_F(session_test2, DISABLED_additional_shutdown_handlers_can_be_added_to_session)
```
isabelsavannah: What's going on here?

pbft/pbft.cpp
```diff
@@ -352,21 +352,35 @@ pbft::forward_request_to_primary(const bzn_envelope& request_env)
...
-    if (msg.has_request() && !op->has_request())
+    if (!op->has_request())
     {
-        if (this->crypto->hash(msg.request()) != msg.request_hash())
+        if (msg.request_type() == PBFT_REQUEST_PAYLOAD)
         {
-            LOG(info) << "Not recording request because its hash does not match";
-            return;
+            assert(env.piggybacked_requests_size()); // we must have at least 1 request
+            if (this->crypto->hash(env.piggybacked_requests(0)) != msg.request_hash())
```
isabelsavannah: It may be simpler for this check to be outside the scope of this method - since the caller likely already identified the pbft_operation it passed to us by way of matching the hashes

pbft/pbft.cpp
```diff
@@ -352,21 +352,35 @@ pbft::forward_request_to_primary(const bzn_envelope& request_env)
...
+            {
+                LOG(info) << "Not recording request because hashes do not match";
+                return;
+            }
+            op->record_request(env.piggybacked_requests(0));
```
isabelsavannah: We only record the first one?

pbft/pbft.cpp
```diff
@@ -379,14 +393,16 @@ pbft::handle_preprepare(const pbft_msg& msg, const bzn_envelope& original_msg)
...
     }
     else
     {
         auto op = this->operation_manager->find_or_construct(msg, this->current_peers_ptr());
-        op->record_pbft_msg(msg, original_msg);
-        this->maybe_record_request(msg, op);
+        auto env_copy{original_msg};
+        env_copy.clear_piggybacked_requests();
+        op->record_pbft_msg(msg, env_copy);
+        this->maybe_record_request(msg, original_msg, op);
```
isabelsavannah: So the decision here is to not look at piggybacked requests on prepares/commits/newviews, yes? That is fine; I just want to clarify.

pbft/pbft.cpp
```diff
@@ -352,21 +352,35 @@ pbft::forward_request_to_primary(const bzn_envelope& request_env)
     this->failure_detector->request_seen(req_hash);
 }
 
+
 void
-pbft::maybe_record_request(const pbft_msg& msg, const std::shared_ptr<pbft_operation>& op)
+pbft::maybe_record_request(const pbft_msg& msg, const bzn_envelope& env, const std::shared_ptr<pbft_operation>& op)
 {
-    if (msg.has_request() && !op->has_request())
```
isabelsavannah: Are we somehow getting the request body for internal requests as well here?

pbft/pbft.cpp
```diff
@@ -1394,17 +1438,80 @@ pbft::save_checkpoint(const pbft_msg& msg)
     }
 }
 
+
+std::map<bzn::hash_t, int>
```
isabelsavannah: May be cleaner to return a map of hashes to references (or std::reference_wrappers, as the case may be...)

pbft/pbft.hpp
```diff
@@ -208,7 +209,7 @@ namespace bzn
         bool move_to_new_configuration(const hash_t& config_hash, uint64_t view);
         bool proposed_config_is_acceptable(std::shared_ptr<pbft_configuration> config);
 
-        void maybe_record_request(const pbft_msg& msg, const std::shared_ptr<pbft_operation>& op);
+        void maybe_record_request(const pbft_msg& msg, const bzn_envelope& env, const std::shared_ptr<pbft_operation>& op);
```
isabelsavannah: Does this really need to take both a msg and envelope?

pbft/pbft.cpp
```diff
@@ -352,21 +352,35 @@ pbft::forward_request_to_primary(const bzn_envelope& request_env)
...
+            if (this->crypto->hash(msg.request()) != msg.request_hash())
+            {
+                LOG(info) << "Not recording request because hashes do not match";
+                return;
+            }
+            op->record_request(msg.request());
         }
-
-        op->record_request(msg.request());
     }
```
isabelsavannah: Upon further reading, it seems to me as if this method is used in two ways
- saving the request that is part of some pbft message corresponding to an internal request
- saving the (first only) request piggybacked to a preprepare corresponding to a payload request

And that further these cases are invoked from different contexts and share little if any code. Should they not just be different methods, then?

pbft/pbft.hpp
rnistuk: I will review maybe_record_request and see about cleaning it up.

pbft/pbft.cpp
rnistuk: Reviewed with Isabel.

pbft/pbft.cpp
rnistuk: Reviewed with Isabel.

pbft/pbft.cpp
rnistuk: I believe that is correct.

pbft/pbft.hpp
rnistuk: Reviewed and rewritten with Isabel

node/test/session_test.cpp
rnistuk: This has been corrected. Thanks Isabel.

PR #335, opened 2019-06-24
KEP-1523 Swarmdb ESR parsing mismatch for http_port

rnistuk:
This is an emergency bug fix. The invalid unit test code has been commented out pending changes to the solidity side.


utils/esr_peer_info.cpp
```diff
@@ -249,69 +249,100 @@ namespace
...
                 }
                 break;
-                case NODE_HOST: {
+
+                case NODE_HOST:
+                {
                     host = hex_to_char_string(line);
                     trim_right_nulls(host);
-                    state = NA_3;
+                    assert(text_size == host.size());
```
ebruck: you sure you want asserts?

PR #338, opened 2019-06-28
KEP-1538 Modified ESR Get Peer Info Init test to use new contract, fi…

rnistuk:
…xed get peer info parser to read node names greater than 32 characters

### DIFF
--- a/utils/esr_peer_info.cpp
+++ b/utils/esr_peer_info.cpp
@@ -321,19 +321,18 @@ namespace
                         LOG(warning) << "Invalid value for node name string length:[" << text_size << "]";
                     }
                     state = NODE_NAME;
+                    name = "";
                 }
                 break;
 
                 case NODE_NAME:
                 {
-                    name = hex_to_char_string(line);
+                    name += hex_to_char_string(line);
                     trim_right_nulls(name);
-                    assert(text_size == name.size());
-                    if (name.empty())
+                    if (text_size == name.size())
                     {
-                        LOG(warning) << "Empty value for host name, node may not exist";
+                        state = FINISHED;
                     }
-                    state = FINISHED;
                 }
                 break;
 
--- a/utils/esr_peer_info.cpp
+++ b/utils/esr_peer_info.cpp
@@ -324,13 +324,13 @@ namespace
                         LOG(warning) << "Invalid value for node name string length:[" << text_size << "]";
                     }
                     state = NODE_NAME;
-                    name = "";
+                    name.clear();
                 }
                 break;
 
                 case NODE_NAME:
                 {
-                    name += hex_to_char_string(line);
+                    name.append(hex_to_char_string(line));
                     trim_right_nulls(name);
                     if (text_size == name.size())
                     {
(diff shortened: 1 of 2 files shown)


utils/esr_peer_info.cpp
```diff
@@ -321,19 +321,18 @@ namespace
...
                     }
                     state = NODE_NAME;
+                    name = "";
                 }
                 break;
 
                 case NODE_NAME:
                 {
-                    name = hex_to_char_string(line);
+                    name += hex_to_char_string(line);
```
ebruck: why +=?

utils/esr_peer_info.cpp
```diff
@@ -321,19 +321,18 @@ namespace
                         LOG(warning) << "Invalid value for node name string length:[" << text_size << "]";
                     }
                     state = NODE_NAME;
+                    name = "";
```
ebruck: name.clear()

utils/esr_peer_info.cpp
rnistuk: I can change it to append.

utils/esr_peer_info.cpp
rnistuk: cool, using clear.

utils/esr_peer_info.cpp
```diff
@@ -408,14 +410,19 @@ namespace
...
+
         return std::string{"0x"
             + pad_str_to_mod_64(GET_PEER_INFO_SIGNATURE)
             + pad_str_to_mod_64("00000040")                 // first param type?
-            + pad_str_to_mod_64("00000080")                 // second param type?
-            + size_type_to_hex(swarm_id.size())             // size of swarm id string (pre hexification)
-            + pad_str_to_mod_64(string_to_hex(swarm_id))    // parameter 1 - swarm id
-            + size_type_to_hex(peer_id.size(), 64)          // size of peer id (pre hexification)
-            + pad_str_to_mod_64(string_to_hex(peer_id))     // parameter 2 - peer id
+            + size_type_to_hex(PARAMS.size() / 2)           // size of params blob in bytes, padded to 8 chars
```
nickroeker: I _think_ there's an issue here if I'm understanding correctly. This will always indicate the start of the value for the second parameter (`peer_id`) is always halfway past the size of the encoding of _both_ parameters. Since these two strings can vary in size independently, this will not always be accurate.

Since the first parameter is a string, 32 bytes are used to indicate the offset to find the encoding for it. As the first parameter, it will be at the end of the parameter list. The second and last parameter is also a string, so another 32 bytes. The real data therefore begins 64 bytes later, which is where the `00000040` (or 0x40, or 64) comes from in the first bit. The mod64 is of course because hex chars are half bytes. The first string will always start here.

The next 32 bytes needs to indicate the offset for the second parameter's encoding. The varies based on the encoding (32 byte length, n*32 byte data value) of the first parameter, and only the first parameter. `PARAMS.size() / 2` accounts for the size of both parameters, which is incorrect.

utils/esr_peer_info.cpp
```diff
@@ -408,14 +410,19 @@ namespace
         static const auto PEER_INFO_ABI{str_to_json(GET_PEER_INFO_ABI)};
         static const auto GET_PEER_INFO_SIGNATURE{PEER_INFO_ABI["signature"].asCString() + 2};
 
+        static const std::string PARAMS
```
nickroeker: Fairly certain this shouldn't be `static`?

PR #343, opened 2019-07-04
KEP-1564 Some compile options missing in openssl

ebruck:


ebruck:
missed build from source cmake

PR #347, opened 2019-07-16
KEP-1606 - don't forward messages with invalid timestamps

paularchard:
Fix for issue on testnet where bad client timestamp was causing viewchanges to occur.

### DIFF
--- a/pbft/pbft.cpp
+++ b/pbft/pbft.cpp
@@ -301,8 +301,32 @@ pbft::setup_request_operation(const bzn_envelope& request_env, const bzn::hash_t
 }
 
 void
+pbft::send_error_response(const bzn_envelope& request_env, const std::shared_ptr<session_base>& session
+    , const std::string& msg) const
+{
+    database_msg req;
+    if (session && req.ParseFromString(request_env.database_msg()))
+    {
+        database_response resp;
+        *resp.mutable_header() = req.header();
+        resp.mutable_error()->set_message(msg);
+
+        session->send_message(std::make_shared<std::string>(this->wrap_message(resp).SerializeAsString()));
+    }
+}
+
+void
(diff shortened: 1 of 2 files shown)


pbft/pbft.cpp
```diff
@@ -303,6 +303,25 @@ pbft::setup_request_operation(const bzn_envelope& request_env, const bzn::hash_t
 void
 pbft::handle_request(const bzn_envelope& request_env, const std::shared_ptr<session_base>& session)
 {
+    if (request_env.timestamp() < (this->now() - MAX_REQUEST_AGE_MS) || request_env.timestamp() > (this->now() + MAX_REQUEST_AGE_MS))
+    {
+        database_msg req;
+        if (session && req.ParseFromString(request_env.database_msg()))
+        {
+            database_response resp;
+            resp.set_allocated_header(new database_header(req.header()));
```
ebruck: you don't need to do this.
resp.mutable_header()->xxxx will allocate automatically. We no longer use set_allocated in the daemon.

pbft/pbft.cpp
```diff
@@ -303,6 +303,25 @@ pbft::setup_request_operation(const bzn_envelope& request_env, const bzn::hash_t
...
+            auto err = new database_error;
+            err->set_message(TIMESTAMP_ERROR_MSG);
+            resp.set_allocated_error(err);
```
ebruck: same here

pbft/pbft.cpp
paularchard: How do I set the entire header from another header?

pbft/pbft.cpp

paularchard:
Also added a send error response to the "too large request" case as that would also cause a re-transmit loop otherwise.

pbft/pbft.cpp
ebruck: *response.mutable_header() = request.header();

PR #350, opened 2019-07-22
KEP-1428 Stabiity changes to fix node crashing after view changes

rnistuk:


proto/pbft.proto
```diff
@@ -102,6 +102,7 @@ message pbft_membership_msg
     // for set_state
     bytes state_data = 5;
     bzn_envelope newview_msg = 6;
+    bytes current_configuration = 8; // pass the current_config as bytes in pbft handle gete
```
paularchard: comment is mangled

rnistuk: Thanks Paul, I'll fix that.

PR #353, opened 2019-07-26
KEP-1416 macOS only - flush the database during successful call to create and update, thi…

rnistuk:
…s ensures that the SIGABRT is avoided

### DIFF
--- a/storage/rocksdb_storage.cpp
+++ b/storage/rocksdb_storage.cpp
@@ -102,10 +102,23 @@ rocksdb_storage::create(const bzn::uuid_t& uuid, const bzn::key_t& key, const bz
         this->update_metadata_size(uuid, NAMESPACE_KEY, SIZE_KEY, ns_prev_size + value.size() + key.size());
         this->update_metadata_size(uuid, SIZE_KEY, key, value.size() + key.size());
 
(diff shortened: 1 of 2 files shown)


storage/rocksdb_storage.cpp
```diff
@@ -158,6 +171,10 @@ rocksdb_storage::update(const bzn::uuid_t& uuid, const bzn::key_t& key, const bz
         this->update_metadata_size(uuid, NAMESPACE_KEY, SIZE_KEY, ns_prev_size - prev_size + value.size() + key.size());
         this->update_metadata_size(uuid, SIZE_KEY, key, value.size() + key.size());
 
+#ifdef __APPLE__
```
ebruck: perhaps move the flush to the update_metadata_size instead?

rnistuk: That's a good idea.

ebruck: Let's create a backlog item to properly understand what is going on here.

rnistuk: After discussions with Ed, I've decided the leave the flushes where they are and create a backlog item to find the real problem.

PR #355, opened 2019-08-06
KEP-1618 - don't add requests to newview if we don't have them

paularchard:
Fix for testnet-2 crash July 18 2019

### DIFF
--- a/pbft/pbft.cpp
+++ b/pbft/pbft.cpp
@@ -1307,7 +1307,7 @@ pbft::is_valid_newview_message(const pbft_msg& theirs, const bzn_envelope& origi
         viewchange_envelopes_from_senders[viewchange_env.sender()] = viewchange_env;
     }
 
-    pbft_msg ours = this->build_newview(theirs.view(), viewchange_envelopes_from_senders);
+    pbft_msg ours = this->build_newview(theirs.view(), viewchange_envelopes_from_senders, false);
     if (ours.pre_prepare_messages_size() != theirs.pre_prepare_messages_size())
     {
         LOG(error) << "is_valid_newview_message - expected " << ours.pre_prepare_messages_size()
@@ -1425,7 +1425,8 @@ pbft::make_newview(
 }
 
 pbft_msg
-pbft::build_newview(uint64_t new_view, const std::map<uuid_t,bzn_envelope>& viewchange_envelopes_from_senders) const
+pbft::build_newview(uint64_t new_view, const std::map<uuid_t,bzn_envelope>& viewchange_envelopes_from_senders
+    , bool attach_reqs) const
 {
     //  Computing O (set of new preprepares for new-view message)
     std::map<uint64_t, bzn_envelope> pre_prepares;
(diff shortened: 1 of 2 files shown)


pbft/pbft.cpp
```diff
@@ -1307,7 +1307,7 @@ pbft::is_valid_newview_message(const pbft_msg& theirs, const bzn_envelope& origi
         viewchange_envelopes_from_senders[viewchange_env.sender()] = viewchange_env;
     }
 
-    pbft_msg ours = this->build_newview(theirs.view(), viewchange_envelopes_from_senders);
+    pbft_msg ours = this->build_newview(theirs.view(), viewchange_envelopes_from_senders, false);
     if (ours.pre_prepare_messages_size() != theirs.pre_prepare_messages_size())
```
rnistuk: Cool, while checking for new view message validity we do not need to worry about adding requests we don't have.

pbft/pbft.cpp
```diff
@@ -1470,7 +1471,7 @@ pbft::build_newview(uint64_t new_view, const std::map<uuid_t,bzn_envelope>& view
             }
 
             auto env = this->wrap_message(pre_prepare);
-            if (pre_prepare.request_type() == pbft_request_type::PBFT_REQUEST_PAYLOAD)
+            if (pre_prepare.request_type() == pbft_request_type::PBFT_REQUEST_PAYLOAD && attach_reqs)
```
rnistuk: OK, here's where we attach the requests...

pbft/pbft.cpp
```diff
@@ -1480,7 +1481,15 @@ pbft::build_newview(uint64_t new_view, const std::map<uuid_t,bzn_envelope>& view
                 }
                 auto op = this->operation_manager->find_or_construct(old_view, pre_prepare.sequence()
                     , pre_prepare.request_hash(), config->get_peers());
```
rnistuk: Cool, checking that the operation has a request before attempting to access it.

pbft/pbft.hpp
```diff
@@ -229,7 +229,7 @@ namespace bzn
         void initiate_viewchange();
         std::shared_ptr<bzn_envelope> make_viewchange(uint64_t new_view, uint64_t n, const std::unordered_map<bzn::uuid_t, std::string>& stable_checkpoint_proof, const std::map<uint64_t, std::shared_ptr<bzn::pbft_operation>>& prepared_operations);
         pbft_msg make_newview(uint64_t new_view_index,  const std::map<uuid_t,bzn_envelope>& viewchange_envelopes_from_senders, const std::map<uint64_t, bzn_envelope>& pre_prepare_messages) const;
-        pbft_msg build_newview(uint64_t new_view, const std::map<uuid_t,bzn_envelope>& viewchange_envelopes_from_senders) const;
+        pbft_msg build_newview(uint64_t new_view, const std::map<uuid_t,bzn_envelope>& viewchange_envelopes_from_senders, bool attach_reqs = false) const;
```
rnistuk: Based on the usage pattern of build_newview, that is the only time attach_reqs is actually set, it is set to false, is false the correct default here?

pbft/pbft.cpp
paularchard: No the default should be true. Changing it.

pbft/pbft.hpp
rnistuk: Cool! Fixed it. Thanks.

PR #357, opened 2019-08-27
KEP-1679 Release tagging no longer works on swarmDB master

ebruck:


.travis.yml
```diff
@@ -120,6 +120,7 @@ after_success:
       fi
     fi
     if [ "$TRAVIS_BRANCH" = "master" ] ; then
+      git lfs fetch --all
```
nickroeker, line 123: Quick question (likely missing context)
If we didn't have the pre-builts from LFS before this, shouldn't the build have failed prior to reaching this point? Or is it building dependencies every time?

ebruck: It pulled what we needed to build, but apparently the pushing to "origin-swarm" needs the refs.

ebruck: git remote add origin-swarm https://${GITHUB_TOKEN}@github.com/bluzelle/swarmdb.git > /dev/null 2>&1
      git push --quiet --set-upstream origin-swarm ${TRAVIS_BRANCH} --tags

nickroeker (approved):

PR #358, opened 2019-08-29
KEP-1702 - various PBFT stability fixes

paularchard:
A mixed bag of bug fixes and improvements to make a busy swarm more stable. More work is in progress. This is an interim commit.

VIEWCHANGE and NEWVIEW messages were being sent repeatedly for the same view number every time there was a failure detected. These messages are very expensive to generate and process. We should now only send once per view.

VIEWCHANGE messages were sometimes being sent for the wrong view number due to a bug in the logic which is now fixed.

Missing pre-prepare messages, which are filled in with a “null database message” were being signed by the nodes that generated them, causing signature mismatches leading to invalid NEWVIEWs, causing stalls. These messages are no longer signed.

Delayed state transfer when we determine we are behind a checkpoint was incorrectly sending state requests when it didn’t need to. This has been fixed.

The point at which incoming messages are verified has been moved to later in the message processing so that messages we end up dropping (duplicate requests, invalid view messages etc., are not verified as that is an expensive process and we were wasting a lot of CPU cycles.

The upper bound on incoming message ids that we will process has been removed. Alternatively it could be set to a much larger number but I’m not sure what would be safe. This was found to have a major impact on stability in parallel request cases as we were frequently exceeding the message window and dropping valid messages, leading to unexecuted requests and view change loops. This is one of the biggest fixes in terms of stability.

The window for acceptable timestamps on requests has been increased substantially. Previously we were frequently not processing requests and their retries quickly enough to stay within the window, causing unexecuted requests and viewchange loops.

The failure handler has been modified to:

have a longer delay before it initially triggers a failure

have a secondary delay value (larger) that it uses after an initial failure, to allow time for a viewchange to occur and not immediately trigger another failure.

allow both the initial timeout value and the secondary timeout value to be configurable.

If the local view is invalid, don’t record any requests forwarded to the primary with the failure detector. Previously we would record the request, reject all the prepare/commit messages because the view was invalid, and then trigger a failure because we didn’t execute the request.

### DIFF
--- a/node/node.cpp
+++ b/node/node.cpp
@@ -142,12 +142,6 @@ node::priv_protobuf_handler(const bzn_envelope& msg, std::shared_ptr<bzn::sessio
         return;
     }
 
-    if ((!msg.sender().empty()) && (!this->crypto->verify(msg)))
-    {
-        LOG(error) << "Dropping message with invalid signature: " << msg.ShortDebugString().substr(0, MAX_MESSAGE_SIZE);
-        return;
-    }
-
     if (auto it = this->protobuf_map.find(msg.payload_case()); it != this->protobuf_map.end())
     {
         it->second(msg, std::move(session));
--- a/node/test/node_test.cpp
+++ b/node/test/node_test.cpp
@@ -177,6 +177,7 @@ namespace  bzn
         ASSERT_FALSE(node->register_for_message(bzn_envelope::kDatabaseMsg, [](const auto&, auto){}));
     }
 
+#if 0
     TEST(node, test_that_wrongly_signed_messages_are_dropped)
     {
         auto mock_chaos = std::make_shared<NiceMock<bzn::mock_chaos_base>>();
@@ -211,4 +212,5 @@ namespace  bzn
         node->priv_protobuf_handler(anon_msg, mock_session);
         EXPECT_EQ(callback_execute, 1u);
     }
+#endif
 } // namespace bzn
--- a/options/simple_options.cpp
+++ b/options/simple_options.cpp
@@ -88,6 +88,12 @@ simple_options::build_options()
                 (WS_IDLE_TIMEOUT.c_str(),
                         po::value<uint64_t>()->default_value(300000),
                         "websocket idle timeout (ms)")
+                (FD_OPER_TIMEOUT.c_str(),
+                        po::value<uint64_t>()->default_value(30000),
+                        "failure-detector operation timeout (ms)")
+                (FD_FAIL_TIMEOUT.c_str(),
+                        po::value<uint64_t>()->default_value(300000),
+                        "failure-detector second failure timeout (ms)")
                 (SWARM_INFO_ESR_ADDRESS.c_str(),
                         po::value<std::string>()->default_value(bzn::utils::DEFAULT_SWARM_INFO_ESR_ADDRESS),
                         "Address of ESR Swarm Info contract")
--- a/pbft/pbft.cpp
+++ b/pbft/pbft.cpp
@@ -196,6 +196,12 @@ pbft::handle_membership_message(const bzn_envelope& msg, std::shared_ptr<bzn::se
         return;
     }
 
+    if ((!msg.sender().empty()) && (!this->crypto->verify(msg)))
+    {
+        LOG(error) << "Dropping message with invalid signature: " << msg.ShortDebugString().substr(0, MAX_MESSAGE_SIZE);
+        return;
+    }
+
     const auto hash = this->crypto->hash(msg);
 
     std::lock_guard<std::mutex> lock(this->pbft_lock);
@@ -230,6 +236,12 @@ pbft::handle_message(const pbft_msg& msg, const bzn_envelope& original_msg)
 
     if (!this->preliminary_filter_msg(msg))
     {
+        return;
+    }
+
+    if ((!original_msg.sender().empty()) && (!this->crypto->verify(original_msg)))
+    {
+        LOG(error) << "Dropping message with invalid signature: " << original_msg.ShortDebugString().substr(0, MAX_MESSAGE_SIZE);
         return;
     }
 
@@ -277,13 +289,6 @@ pbft::preliminary_filter_msg(const pbft_msg& msg)
         {
             LOG(debug) << boost::format("Dropping message because sequence number %1% <= %2%") % msg.sequence()
                 % this->get_low_water_mark();
-            return false;
-        }
-
-        if (msg.sequence() > this->get_high_water_mark())
-        {
-            LOG(debug) << boost::format("Dropping message because sequence number %1% greater than %2%") % msg.sequence()
-                % this->get_high_water_mark();
             return false;
         }
     }
@@ -367,6 +372,13 @@ pbft::handle_request(const bzn_envelope& request_env, const std::shared_ptr<sess
         LOG(debug) << "Rejecting duplicate request: " << request_env.ShortDebugString().substr(0, MAX_MESSAGE_SIZE);
         return;
     }
+
+    if ((!request_env.sender().empty()) && (!this->crypto->verify(request_env)))
+    {
+        LOG(error) << "Dropping message with invalid signature: " << request_env.ShortDebugString().substr(0, MAX_MESSAGE_SIZE);
+        return;
+    }
+
     this->saw_request(request_env, hash);
     auto op = setup_request_operation(request_env, hash);
     this->do_preprepare(op);
@@ -381,7 +393,10 @@ pbft::forward_request_to_primary(const bzn_envelope& request_env)
     const bzn::hash_t req_hash = this->crypto->hash(request_env);
     LOG(info) << "Forwarded request to primary, " << bzn::bytes_to_debug_string(req_hash);
 
-    this->failure_detector->request_seen(req_hash);
+    if (this->is_view_valid())
+    {
+        this->failure_detector->request_seen(req_hash);
+    }
 }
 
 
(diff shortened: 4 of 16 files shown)


node/node.cpp
```diff
@@ -142,12 +142,6 @@ node::priv_protobuf_handler(const bzn_envelope& msg, std::shared_ptr<bzn::sessio
         return;
     }
 
-    if ((!msg.sender().empty()) && (!this->crypto->verify(msg)))
```
rnistuk, line 145: cool, we've stopped doing signatures in this case

node/test/node_test.cpp
```diff
@@ -177,6 +177,7 @@ namespace  bzn
         ASSERT_FALSE(node->register_for_message(bzn_envelope::kDatabaseMsg, [](const auto&, auto){}));
     }
 
+#if 0
```
rnistuk, line 180: Shouldn't we just remove this test?

options/simple_options.cpp
```diff
@@ -88,6 +88,12 @@ simple_options::build_options()
                 (WS_IDLE_TIMEOUT.c_str(),
                         po::value<uint64_t>()->default_value(300000),
                         "websocket idle timeout (ms)")
+                (FD_OPER_TIMEOUT.c_str(),
```
rnistuk, line 91: Cool, setting defaults for FD_OPER_TIMEOUT AND FD_FAIL_TIMEOUT. It's good that we are providing this control to dev ops.

pbft/pbft.cpp
```diff
@@ -233,6 +239,12 @@ pbft::handle_message(const pbft_msg& msg, const bzn_envelope& original_msg)
         return;
     }
 
+    if ((!original_msg.sender().empty()) && (!this->crypto->verify(original_msg)))
```
rnistuk, line 242: Little bit of duplication here. Similar to pfft.cpp:199. Probably not worth changing...

pbft/pbft.cpp
```diff
@@ -280,7 +292,8 @@ pbft::preliminary_filter_msg(const pbft_msg& msg)
             return false;
         }
 
-        if (msg.sequence() > this->get_high_water_mark())
+        // FORNOW: don't test for high water mark, too easy for us to get out of range
+        if (0) //msg.sequence() > this->get_high_water_mark())
```
rnistuk: Should we just remove this block?

pbft/pbft.cpp
```diff
@@ -367,6 +380,13 @@ pbft::handle_request(const bzn_envelope& request_env, const std::shared_ptr<sess
         LOG(debug) << "Rejecting duplicate request: " << request_env.ShortDebugString().substr(0, MAX_MESSAGE_SIZE);
         return;
     }
+
+    if ((!request_env.sender().empty()) && (!this->crypto->verify(request_env)))
```
rnistuk, line 376: More duplication.

pbft/pbft.cpp
```diff
@@ -991,20 +1014,30 @@ pbft::handle_failure()
 }
 
 void
-pbft::initiate_viewchange()
+pbft::initiate_viewchange(std::optional<uint64_t> opt_view)
```
rnistuk, line 1009: Sweet! std::optional!

pbft/pbft.cpp
```diff
@@ -991,20 +1014,30 @@ pbft::handle_failure()
...
 {
-    pbft_msg view_change;
-    std::vector<bzn_envelope> requests;
+    uint64_t view_to_set = opt_view.has_value() ? *opt_view : this->view.value() + 1;
+    if (view_to_set > this->last_view_sent.value())
```
rnistuk, line 1012: OK, cool, only bother if the view is greater than the current view.

pbft/pbft.hpp
```diff
@@ -41,8 +41,8 @@ namespace
     const std::chrono::seconds NEW_CONFIG_INTERVAL{std::chrono::seconds(30)};
     const std::chrono::seconds JOIN_RETRY_INTERVAL{std::chrono::seconds(30)};
     const uint64_t CHECKPOINT_INTERVAL = 100; //TODO: KEP-574
-    const double HIGH_WATER_INTERVAL_IN_CHECKPOINTS = 2.0; //TODO: KEP-574
-    const uint64_t MAX_REQUEST_AGE_MS = 300000; // 5 minutes
+    const double HIGH_WATER_INTERVAL_IN_CHECKPOINTS = 200.0; //TODO: KEP-574
```
rnistuk, line 44: This is a large increase in for the HIGH_WATER_INTERVAL_IN_CHECKPOINTS value, just checking...

pbft/pbft.hpp
```diff
@@ -41,8 +41,8 @@ namespace
...
+    const uint64_t MAX_REQUEST_AGE_MS = 3600000; // 1 hour
```
rnistuk, line 45: OK, we'll keep trying requests for up to an hour.

pbft/pbft.hpp
```diff
@@ -226,7 +226,7 @@ namespace bzn
         void join_swarm();
 
         // VIEWCHANGE/NEWVIEW Helper methods
-        void initiate_viewchange();
+        void initiate_viewchange(std::optional<uint64_t> opt_view = std::nullopt);
```
rnistuk, line 229: Ah good, std::nullopt is the default.

pbft/pbft_failure_detector.cpp
```diff
@@ -15,39 +15,45 @@
...
+        , options(options)
         , request_progress_timer(this->io_context->make_unique_steady_timer())
 {
 }
 
 void
 pbft_failure_detector::start_timer()
 {
-    this->request_progress_timer->expires_from_now(operation_timeout);
+    this->request_progress_timer->expires_from_now(this->options->get_fd_oper_timeout());
```
rnistuk, line 31: cool, propagating the new options...

pbft/pbft_failure_detector.cpp
```diff
@@ -15,39 +15,45 @@
...
     this->request_progress_timer->async_wait(std::bind(&pbft_failure_detector::handle_timeout, shared_from_this(), std::placeholders::_1));
 }
 
 void
-pbft_failure_detector::handle_timeout(boost::system::error_code /*ec*/)
+pbft_failure_detector::handle_timeout(boost::system::error_code ec)
 {
+    if (ec)
```
rnistuk, line 38: Should there be a log message for this?

pbft/test/pbft_test.cpp
```diff
@@ -225,7 +225,7 @@ namespace bzn::test
         service.apply_operation(op);
     }
 
-    TEST_F(pbft_test, messages_after_high_water_mark_dropped)
+    TEST_F(pbft_test, DISABLED_messages_after_high_water_mark_dropped)
```
rnistuk: Should we try to make this test work?

rnistuk (commented):
Builds cleanly
all unit tests pass
I think the changes are good, but I shall defer to Isabel's review on this one.

node/test/node_test.cpp
paularchard: There are a couple of tests that need to be updated based on the validation changes. I deferred that for now as I'm still hunting down problems.

pbft/pbft.cpp
paularchard: I'm still debating whether to remove or make the window big. Was awaiting feedback.

pbft/pbft.hpp
paularchard: See above

pbft/pbft_failure_detector.cpp
paularchard: It's a cancel in general, which we don't need to log. I guess I could do an explicit test.

pbft/test/pbft_test.cpp
paularchard: See above

pbft/pbft.cpp
isabelsavannah: Might as well just remove it. It only exists as prevention of a (relatively niche) attack that could be performed by a byzantine primary.

pbft/pbft.cpp
isabelsavannah: Consider changing the name of this method to communicate that it may not actually initiate a viewchange

pbft/pbft_failure_detector.cpp
```diff
@@ -15,39 +15,45 @@
...
+    {
+        return;
+    }
+
     std::lock_guard<std::mutex> lock(this->lock);
 
+    assert(!this->ordered_requests.empty());
```
isabelsavannah, line 45: This is intentionally a tamer solution (as opposed to the more drastic - stop caring about request ordering at all, only require progress), correct?

node/node.cpp
isabelsavannah: Is pbft the only consumer of this interface that depends on the verification here? I am willing to believe that it is, I'm just not sure off hand.

pbft/test/pbft_test.cpp
isabelsavannah: Shouldn't it still work per the new, very high high water mark? Or be deleted entirely if the HWM is removed.

isabelsavannah (approved):

pbft/pbft_failure_detector.cpp
paularchard: Actually this is a leftover from some changes I was playing with. The request ordering change didn't show any clear improvements and caused some other complications that made it less desirable.

PR #360, opened 2019-09-24
KEP-1365 Swarm crashes when using non-utf8 data

ebruck:


rnistuk (approved):
Looks good, bytes instead of strings in the protobuf messages.

PR #362, opened 2019-09-27
KEP-1593: Replace dynamic peering with simpler approach

isabelsavannah:
- Implement a new module "peers_beacon", which is responsible for always knowing what the current peers list
- peers_beacon decides which source (esr/url/filesystem) to get peers from based on config, and refreshes at (configurable) interval
- Remove the old bootstrap_peers module, as it is now redundant. Some of it's code has been lifted into peers_beacon.
- Add a mockable interface to utils in order to test peers_beacon's usage of the esr/url based peer fetch methods. The analogous code in bootstrap_peers was previously not tested (at least not by unit tests)
- remove the old dynamic peering protocol; instead we simply consult the peers beacon any time we need the peers list and changes in the peers list are outside the scope of pbft. Under the assumptions that the peers list is not made to change "too quickly", and that all the nodes in it are honest, the chance of something going wrong as a result should be very small.
- remove the old swarm joining logic (this is distinct from the dynamic peering protocol itself). New nodes simply start participating in the swarm; they will still need a checkpoint to happen before they can do much meaningful. They no longer need a view change also, however. As an additional task, we may want to implement some accelerated catchup protocol that does not require a checkpoint.
- removed the concept of "configurations" for pbft. This was a concept that in theory we could extend to accomplish a couple different goals, but productively doing so was dependant on the well behaviour of the old dynamic peering mechanism.
- The preprepared/prepared/committed status of an operation is now stored explicitly, rather than computed based on stored messages- this is necessary so that operations cannot become un-prepared based on a change in the peers list.
- implement tests for basic pbft operations where the peers list changes at various points during the process (pbft_changeover_tests)
- the identity of the primary is now stored explicitly, rather than computed based on the peers list. The primary only changes when a new-view is performed. This is required to maintain the guarantee that there cannot be two primaries of a single view, even when the peers list might change during a viewchange.
- several modules now have their own pointer to the peers_beacon, rather than expecting their caller to provide the peers list at each call
- pbft now rejects pbft messages from non-peers (this is how leaving the swarm is accomplished, even if the node that is to leave has not actually shut down yet)


pbft/pbft.cpp
```diff
@@ -271,6 +274,7 @@ pbft::handle_message(const pbft_msg& msg, const bzn_envelope& original_msg)
 bool
 pbft::preliminary_filter_msg(const pbft_msg& msg)
 {
+
```
paularchard: (pedantic) Spurious blank line

pbft/pbft.cpp
```diff
@@ -387,8 +391,15 @@ pbft::handle_request(const bzn_envelope& request_env, const std::shared_ptr<sess
...
+    auto primary = this->get_current_primary();
+    if (!primary.has_value())
+    {
+        LOG(error) << "Would forward request to primary, but we don't know the primary so we can't";
+        return;
+    }
+
     this->node
-        ->send_signed_message(bzn::make_endpoint(this->get_primary()), std::make_shared<bzn_envelope>(request_env));
+        ->send_signed_message(bzn::make_endpoint(primary.value()), std::make_shared<bzn_envelope>(request_env));
```
paularchard: Not that we "support" Mac, but .value() seems to cause a problem with Mac. Possibly use *primary instead.

pbft/pbft.cpp
```diff
@@ -907,33 +750,34 @@ pbft::do_committed(const std::shared_ptr<pbft_operation>& op)
...
-
-    if (ec)
-    {
-        LOG(error) << "handle_new_config_timeout error: " << ec.message();
-        return;
-    }
-
-    this->initiate_viewchange();
+    auto primary = this->get_current_primary();
+    return primary.has_value() && primary.value().uuid == this->uuid;
```
paularchard: Not that we "support" Mac, but .value() seems to cause a problem with Mac. Possibly use *primary instead.

pbft/operations/pbft_persistent_operation.hpp
```diff
@@ -17,16 +17,17 @@
...
 #include <proto/pbft.pb.h>
+#include <peers_beacon/peers_beacon_base.hpp>
 
 namespace bzn
 {
     class pbft_persistent_operation : public pbft_operation
     {
     public:
-        pbft_persistent_operation(uint64_t view, uint64_t sequence, const bzn::hash_t& request_hash, std::shared_ptr<bzn::storage_base> storage, size_t );
+        pbft_persistent_operation(uint64_t view, uint64_t sequence, const bzn::hash_t& request_hash, std::shared_ptr<bzn::storage_base> storage, std::shared_ptr<bzn::peers_beacon_base> peers);
```
paularchard: Each operation is storing the peers beacon pointer. This seems a little redundant. Previously a peers pointer was stored but it could be different based on the config. In this case it will always be the same. 

It seems like it is only used in is_ready_for_(commit|execute), which are only called from pbft::maybe_advance_operation. Perhaps it could be passed in from there?

utils/utils_interface_base.hpp
```diff
@@ -0,0 +1,36 @@
...
+#pragma once
+#include <include/bluzelle.hpp>
+#include <string>
+#include <vector>
+#include <peers_beacon/peer_address.hpp>
+
+
+namespace bzn
+{
+    class utils_interface_base
```
paularchard, line 24: We do something similar in bzapi, for the same reason (mockable). It would be nice to consolidate the code. Even better, put all the ESR and HTTP stuff into a single library and share it.

peers_beacon/peers_beacon.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+    std::sort(new_ordered->begin(), new_ordered->end(),
+        [](const auto& peer1, const auto& peer2)
+        {
+            return peer1.uuid.compare(peer2.uuid) < 0;
+        });
+
+    this->internal_current = std::make_shared<const peers_list_t>(new_peers);
+    this->internal_current_ordered = new_ordered;
+
```
paularchard, line 242: I could be wrong but it seems like this isn't thread-safe. The same shared_ptr instance (internal_current, internal_current_ordered) could be written here and read by current() / ordered() simultaneously while the shared_ptr object was in an inconsistent state.

paularchard (changes requested):
In general I like the changes. Much simpler without the config and a peers pointer in the operation :) Plus no join requests!

My only reservation is that we could get into  a situation where we can't validate a viewchange or newview message because the quorum went up since the operation was prepared. Is that a realistic concern?

pbft/operations/pbft_persistent_operation.hpp
isabelsavannah: It could be passed like that, yeah. The only reason I have written it like this is that it seemed more appropriate for the operation to know already, but that's a pretty vauge feeling.

peers_beacon/peers_beacon.cpp
isabelsavannah: I had previously looked at the documentation and concluded that it was thread safe as is, but now that I look more closely I think you're actually right here and it technically should be protected. Shared pointers are apparently highly weird. I'll add a mutex here probably.

isabelsavannah:
My assumption is that the situation where nodes have different peers lists is always transient. Therefore, if a view change fails for that reason, then after some time a new view change will succeed. In fact I am erring on the side of causing that sketchy (in that it happens during the time when the peers lists are not consistent) view change to fail rather than risk it succeeding in some way that violates an invariant (such as two primaries).

Then, for the times when we would process a "stale" viewchange (joining or catch up) if we have to we can just accept it despite not having enough newviews.

peers_beacon/peers_beacon.cpp
```diff
@@ -175,6 +177,8 @@ peers_beacon::fetch_from_esr()
 bool
 peers_beacon::switch_peers_list(const peers_list_t& new_peers)
 {
+    std::unique_lock lock(this->lock);
```
paularchard: This could be moved down to just before where the internal_* members are set.

paularchard (approved):
One slight optimization suggested.

paularchard (approved):

PR #363, opened 2019-10-16
KEP-477 Support for Secure Websockets (WSS)

ebruck:
Since this slides right into the mockable websocket there's no easy way to test the ssl part. An integration test may be the only real way. Unless you have some ideas in nodetest?

There is no peer validation until we figure out what to do there. So, self signed certs do not cause connection failures between nodes in a swarm.

### DIFF
--- a/include/boost_asio_beast.hpp
+++ b/include/boost_asio_beast.hpp
@@ -14,8 +14,12 @@
 
 #pragma once
 
+#include <include/bluzelle.hpp>
 #include <boost/asio.hpp>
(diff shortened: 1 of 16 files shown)


rnistuk (approved):
The code, as usual for Ed, is very beautiful.

There are surprisingly few, well placed changes. 

I fully understood the options part of the changes, and sort of understood the boost::beast related changes.

It looks like the new code passes the existing unit tests. Now on to integration testing.

PR #364, opened 2019-10-23
Task/parchard/kep 1755

paularchard:
You can review them all together against devel, or individually if you prefer.


node/test/node_test.cpp
```diff
@@ -178,13 +178,58 @@ namespace  bzn
...
+    }
+
 #if 0
     TEST(node, test_that_wrongly_signed_messages_are_dropped)
     {
         auto mock_chaos = std::make_shared<NiceMock<bzn::mock_chaos_base>>();
         auto mock_io_context = std::make_shared<NiceMock<bzn::asio::mock_io_context_base>>();
         auto options = std::make_shared<bzn::options>();
-        options->get_mutable_simple_options().set(bzn::option_names::CRYPTO_ENABLED_INCOMING, "true");
+        options->get_mutable_simple_options).set(bzn::option_names::CRYPTO_ENABLED_INCOMING, "true");
```
isabelsavannah: probable typo (admittedly in dead code, but still)

node/node.cpp
```diff
@@ -200,7 +206,13 @@ node::find_session(const boost::asio::ip::tcp::endpoint& ep)
                 , this->options
                 , std::nullopt
                 , this->client_ctx);
-        session->open(this->websocket);
+        session->open(this->websocket, [self = shared_from_this(), ep](const boost::system::error_code& ec)
+        {
+            if (ec && self->error_callback)
+            {
+                self->error_callback(ep, ec);
```
isabelsavannah, line 213: It seems that we only invoke this callback on failures that happen related to the handshake - given that it's being sued to trigger a view change, should we not also call it when the socket fails later, or even when it gracefully closes?

node/node.cpp
```diff
@@ -274,6 +274,50 @@ node::send_signed_message(const bzn::uuid_t& uuid, std::shared_ptr<bzn_envelope>
...
+
+void
+node::multicast_maybe_signed_message(std::shared_ptr<std::vector<boost::asio::ip::tcp::endpoint>> eps, std::shared_ptr<bzn_envelope> msg)
+{
+    this->io_context->post(
+        [self =  shared_from_this(), msg, eps]()
+        {
+            for (const auto& ep : *eps)
+            {
+                self->send_maybe_signed_message(ep, msg);
```
isabelsavannah, line 300: This will compute a signature for each peer, even though that signature is identical.

isabelsavannah (changes requested):

node/node.cpp
paularchard: My assumption was that we would try to reconnect and notice the problem there. I could add a socket error case too. In terms of gracefully closing couldn't a session time out and cause that legitimately?

node/node.cpp
paularchard: The msg is passed as a non-const (shared) pointer down to session::send_signed_message, which signs it if the signature is empty. Shouldn't that do it once and only once?

node/test/node_test.cpp
paularchard: d'oh. Fixed

node/node.cpp
isabelsavannah: Ah, that's fair. In the case where a send fails and we reconnect, it might not be until the next message that we do so, but that's fine.

node/node.cpp
isabelsavannah: Oh, I had not realized that this wasn't async. Consider making it async, but if you decide not to then I agree that it does avoid duplicating the crypto.

isabelsavannah (approved):

PR #365, opened 2019-10-24
KEP-1784 Fix swarmDB & bzapi compiler warnings for gcc 9.2.1

ebruck:


CMakeLists.txt
```diff
@@ -42,6 +42,8 @@ else()
     if("${CMAKE_BUILD_TYPE}" STREQUAL "Release")
         set(warnings "${warnings} -Wno-maybe-uninitialized")
     endif()
+
+    set(warnings "${warnings} -Wno-deprecated-copy -Wno-sign-compare")
```
rnistuk: Isn't the no-sign-compare a righteous warning?

rnistuk (approved):
I am slightly concerned about turning off the no-sign-compare warning, but I suspect you have a good reason for it.

CMakeLists.txt
ebruck: ya, but sadly there is no way around this as beast I think spews this warning.

PR #369, opened 2019-11-14
KEP-1815 Crypto library coverage low

ebruck:


rnistuk (approved):
Cool, all message types have the sender set correctly, even the bad one.

